PR #36, opened 2010-09-07
Add base translation and German translation

jezdez:
The base translation (English) helps translators that don't know how to the gettext based tools.


jezdez:
Just learned this was my first pull request.

PR #98, opened 2012-03-18
Ability to force lowercase tags. Management commands to convert mixed case to lowercase tags.

shacker:


frewsxcv:
Case insensitive tags was added in https://github.com/alex/django-taggit/pull/325 and is available in the latest version of taggit (0.16.0)

PR #104, opened 2012-08-25
Fixing issue when subclassing models

ryanbagwell:
This change creates related names that contain the class name, so that related names won't clash when subclassing taggit models.


frewsxcv:
Closing due to inactivity.

Issue #84, opened 2011-09-28
Cannot resolve keyword 'tagged_items' into field. Choices are: geoloc, id, lat, long


florinmatei:
I have exactly the same issue as the one described here:
http://stackoverflow.com/questions/6560287/django-taggit-deep-relationship-query

I'm getting the error "Cannot resolve keyword 'tagged_items' into field. Choices are: geoloc, id, lat, long".
Any ideas how to solve this issue?

PR #126, opened 2013-04-24
Don't return extra fields for joins resolved through other managers.

joestump:
This *might* be a fix for #84. I had a query that required me to do `taggedmodel__tags__name__in` in my filters, which would trigger this `extra_filters()` bug in #84. My fix simply says, "Is the first piece of this filter me?" If the answer is no, then `extra_filters()` is entirely skipped.

I'm not well versed in the ORM internals so I can't be 100% sure this is a firm fix. What I can say is that it's likely a safer fix than entirely removing `extra_filters()`. I also don't have a test for this; sorry about that!

### DIFF
--- a/taggit/managers.py
+++ b/taggit/managers.py
@@ -129,7 +129,7 @@ class TaggableManager(RelatedField):
         return self.through._meta.db_table
 
     def extra_filters(self, pieces, pos, negate):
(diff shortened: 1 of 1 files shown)


frewsxcv:
Closing due to inactivity.

PR #232, opened 2014-04-25
Add option to override default tag string parsing

### LINKED ISSUE #215: Tag delimiter proposal
Would it be acceptable to suggest a better way of delimiting tags?  Right now the  logic  that determines how a tag string is pretty complex.  If there's a comma, then the tag string is split based on the commas.  Otherwise, the tag string is split based on whitespace.  So, the tag string `Cued Speech, adaptations` will be parsed as two tags (`Cued Speech` and `adaptations`) whereas `Cued Speech` will be parsed as two tags `Cued` and `Speech`.   We recently encountered this issue on our website where some people were attempting to attach a single tag to an item as `Cued Speech`; however, it kept splitting the tag into two tags.  They were pretty confused because on other items when they attempted to attach multiple tags (e.g. `Cued Speech, adaptations, famous people`), the tag list was properly getting parsed into three tags.

In my opinion this is relatively unintuitive behavior for most users.  I would suggest that simplify the approach by adding a regular expression pattern that can be configured via the settings file (e.g. TAGGIT_TAG_DELIMITER).  One could then define a regular expression that splits on all spaces, except when the word is surrounded by quotation marks (e.g. r`(?: |\"(.*?)\"|'.*?')?`.  Alternatively, if one prefers that the split just occur on commas one could define the pattern accordingly (e.g. r'\s*,\s*' to capture the comma and whitespace around the comma).

This would, however, introduce a backwards-incompatible change.  For example, the `parse_tags` function would no longer change its behavior based on whether it detects commas/quotation marks in the string.  Each website would have to decide what an appropriate tag delimiter is and define it accordingly in the settings file.  

Additional work would have to be done for the reverse direction (e.g. `edit_string_for_tags` would need to know how to build the tag string and separator.  This behavior can presumably also be customized via settings (e.g. TAGGIT_TAG_SEPARATOR and TAGGIT_MULTIWORD_QUOTE).  

If there is interest in implementing this, please let me know and I'd be happy to submit a PR.

bburan:
To provide some flexibility in parsing tag strings and creating tag strings from a list of tags, users can now provide their own functions for performing these operations.  Fixes #215.

This is per the discussion in #215.  Please let me know if there's anything else that needs to be fixed before you'll accept the PR.

### DIFF
--- a/taggit/utils.py
+++ b/taggit/utils.py
@@ -1,6 +1,9 @@
 from __future__ import unicode_literals
 
+from importlib import import_module
+
 from django import VERSION
+from django.conf import settings
 from django.utils import six
 from django.utils.encoding import force_text
 from django.utils.functional import wraps
@@ -13,7 +16,7 @@ def _get_field(model, name):
         return model._meta.get_field(name)
 
 
-def parse_tags(tagstring):
+def _parse_tags(tagstring):
     """
     Parses tag input, with multiple word input being activated and
     delineated by commas and double quotes. Quotes take precedence, so
@@ -102,7 +105,7 @@ def split_strip(string, delimiter=','):
     return [w for w in words if w]
 
 
-def edit_string_for_tags(tags):
+def _edit_string_for_tags(tags):
     """
     Given list of ``Tag`` instances, creates a string representation of
     the list suitable for editing by the user, such that submitting the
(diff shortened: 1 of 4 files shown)


frewsxcv:
Can you do the following: 

```
git fetch origin
git rebase origin/develop
```

which should:

1. base the branch off `develop` instead of `master`
2. pick up this change: https://github.com/alex/django-taggit/pull/373 since this pr is not compatible with python 2.6

frewsxcv:
Respond here when you have rebased so I know to look at this

bburan:
@frewsxcv - rebased.

taggit/utils.py
```diff
@@ -1,7 +1,10 @@
 from __future__ import unicode_literals
 
+from importlib import import_module
+
 from django import VERSION
 from django.utils import six
+from django.conf import settings
```
frewsxcv: Nit: can you alphabetize this?

bburan: Sure. In a new commit or should I squash all the commits into a single one?

frewsxcv:
LGTM. Could you squash your changes into one commit?

bburan:
Alphabetized imports and merged commits.

frewsxcv:
Thanks for your contribution! :tada:

frewsxcv:
Just released 0.18.0 that includes this change

PR #239, opened 2014-06-20
add the min_count parameter to managers.most_common function

shellfly:
If ``min_count`` is given, only tags which have a ``num_times`` greater than or equal to ``min_count`` will be returned.


frewsxcv:
Rebased in https://github.com/alex/django-taggit/pull/400

Issue #70, opened 2011-04-17
Need settings option FORCE_LOWERCASE_TAGS


shacker:
Slugs are lowercased by being passed through slugify, but tag names are not. Thus some users can enter "Music" and others enter "music" - these are functionally identical, but taggit creates two tags "music" and "music_1", and tag listings and clouds of course show both versions, which is confusing. 

Django-tagging has a global settings option to force all lowercase tags and I'd love to see the same in Taggit.

frewsxcv:
Case insensitive tags was added in https://github.com/alex/django-taggit/pull/325 and is available in the latest version of taggit (0.16.0)

PR #263, opened 2014-09-15

qris:
I don't think it makes sense to have tags which differ only in case. It's not even possible on MySQL because it violates the unique constraint on tag name:

This PR adds a new setting, `TAGGIT_CASE_INSENSITIVE`, which causes existing tags to be looked up case insensitively. New tags are added with the supplied case, which will be used from then on. The default behaviour is the same as before without this setting.

### DIFF
--- a/taggit/managers.py
+++ b/taggit/managers.py
@@ -4,6 +4,7 @@ from operator import attrgetter
 
 from django import VERSION
 from django.contrib.contenttypes.models import ContentType
+from django.conf import settings
 from django.db import models, router
 from django.db.models.fields import Field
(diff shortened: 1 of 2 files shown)


frewsxcv:
@qris Can you fix the merge conflicts here? I can merge afterwards

mfeliu:
Hi, I have fixed the merged conflicts in another pull request: https://github.com/alex/django-taggit/pull/325 

Could you merge it please?

ivancrneto:
looks like we still have merge conflicts

frewsxcv:
Case insensitive tags was added in https://github.com/alex/django-taggit/pull/325 and is available in the latest version of taggit (0.16.0)

PR #266, opened 2014-09-19
Fix joins on taggable managers defined on submodels of multi-table inheritance.

richardxia:
In Django >= 1.6, taggable managers which are only defined on the submodel in a multi-table inheritance relationship incorrectly attempt to join on `id` instead of the submodel's primary key, which is its foreign key to the parent model. This is probably best explained by reading the test that I wrote in this branch, but to reproduce the bug, I essentially copied the Pet and HousePet relationship defined for the other tests and defined the TaggableManager only on HousePet. You can confirm that my test fails on Django 1.6 without the fix contained in this branch.

The reason this fails is that the `get_joining_columns()` method on TaggableManager hardcodes `id` and `object_id` for the generic foreign key to and from fields. This method is only used in Django 1.6 and 1.7, so Django 1.5 and 1.4 aren't affected. To fix this, I copied the definition of [`get_joining_columns()` from Django 1.6.7's ForeignObject](https://github.com/django/django/blob/1.6.7/django/db/models/fields/related.py#L1019-L1021), which is what TaggableManager seems to be modeled after, which dynamically determines which columns to use based on the related fields.

I tested this fix by running the tests, including my new tests, on Django 1.4.8, 1.5.4, 1.6.7, and 1.7.

### DIFF
--- a/taggit/managers.py
+++ b/taggit/managers.py
@@ -447,10 +447,8 @@ class TaggableManager(RelatedField, Field):
             return self._get_mm_case_path_info(direct=False)
 
     def get_joining_columns(self, reverse_join=False):
-        if reverse_join:
-            return (("id", "object_id"),)
-        else:
-            return (("object_id", "id"),)
+        source = self.reverse_related_fields if reverse_join else self.related_fields
+        return tuple([(lhs_field.column, rhs_field.column) for lhs_field, rhs_field in source])
 
     def get_extra_restriction(self, where_class, alias, related_alias):
         extra_col = self.through._meta.get_field_by_name('content_type')[0].column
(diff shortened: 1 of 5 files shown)


chhantyal:
Why is this not reviewed? This still issue with Django 1.7, 1.8

If you use model inheritance, you can't use taggit.

chhantyal:
@richardxia Thanks a lot for the patch. I applied to my fork to continue project https://github.com/uhuramedia/django-taggit/commit/9838b9ce97483d494723fb18a6c788ed63a1453c

frewsxcv:
Any chance this could get rebased to fix the merge conflicts? Also, has anyone tested this with Django 1.8?

chhantyal:
@frewsxcv Yes, my fork is working on Django 1.8

vst:
I wanted to apply this patch but the relevant section of the code has changed. Anybody who has a more recent patch for this?

PR #273, opened 2014-10-11
Fix #272 (wrong slugs for non-latin chars)

### LINKED ISSUE #272: Unicode tags do not have correct slugs
When adding greek tags  I get wrong slugs: The first greek tag gets an empty slug while the next ones get _1, _2, _3 etc. This behavior can be fixed by including unidecode and modifying line 79 of models.py from ``slug = default_slugify(tag)`` to ``slug = unidecode.unidecode(default_slugify(tag))``.

Unidecode is a nice library that translates non-latin strings to their equivalent latin ones. For instance,

<pre>
import unidecode
unidecode.unidecode(u'δοκιμή') # 'dokime'
unidecode.unidecode(u'greek ελληνικά') # 'greek ellenika'
</pre>

I have to notice that this (translating non-latin characters to their latin equivalents) is the behavior that django-admin *also* follows when prepopulating fields for slugs.

If needed I can also provide a PR for this.

spapas:
This PR resolves #272.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,4 +41,5 @@ setup(
     ],
     include_package_data=True,
     zip_safe=False,
+    install_requires=install_requires,
 )
--- a/setup.py
+++ b/setup.py
@@ -5,6 +5,10 @@ f = open('README.rst')
 readme = f.read()
 f.close()
 
+install_requires = (
+    "Unidecode>=0.04.14",
+)
+
 setup(
     name='django-taggit',
     version='0.12.2',
--- a/taggit/models.py
+++ b/taggit/models.py
@@ -8,6 +8,8 @@ from django.utils.encoding import python_2_unicode_compatible
 from django.utils.translation import ugettext_lazy as _
 from django.utils.translation import ugettext
 
+from unidecode import unidecode
+
 try:
     from django.contrib.contenttypes.fields import GenericForeignKey
 except ImportError:  # django < 1.7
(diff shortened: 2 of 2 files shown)


yohanboniface:
Is there any official status of this PR from the maintainers?

I'm interested too to have it merged. And I also think that using unidecode only if installed (as suggested in https://github.com/alex/django-taggit/pull/273#issuecomment-66955534) could be a nice way to keep the default installation light and without deps, and to prevent any licence change issue.

frewsxcv:
It'd be great if this could be a feature one could opt-in to instead of having a library that everyone has to install. Maybe some way for the user to specify a function they want to use to generate a slug? Also, I'm not really a fan of adding a GPL dependency to a BSD licensed library.

spapas:
@frewsxcv I proposed on my 15 December 2014 comment (use a try/catch block to check if unidecode is installed and fall back to normal slugify if it's not bee installed) however I don't want to provide a PR until I know it'll be merged :relaxed:

frewsxcv:
Anyone have opinions on #315 before I merge it?

spapas:
@frewsxcv :+1:

frewsxcv:
#315 has been merged

0.15.0 has been released which includes it https://github.com/alex/django-taggit/blob/develop/CHANGELOG.txt#L4

frewsxcv:
Also, if anyone wants to go above and beyond, we'd probably eventually want a regression test and also a mentioning of the optional dependency (unidecode) in the documentation.

PR #289, opened 2015-02-10
add a check for individual tag length in form widget

twig:
Prevents a 500 error if a user enters in a tag longer than 100 characters.

At the moment, it doesn't seem to check if the tag length is valid before saving it.


frewsxcv:
There are merge conflicts

twig:
I've updated the branch, looks like we wrote the same import but had it in a different order.

Issue #284, opened 2014-12-30
related.py gone in 1.8


phoebebright:
See this ticket: https://code.djangoproject.com/ticket/21414

Currently failing when running dev version of 1.8 with error:

No module named related

collinanderson:
Looks like Tim Graham confirmed your idea, but I ran into more and more problems even after that.

collinanderson:
I should mention that things are going to change again after https://github.com/django/django/pull/3114 is merged.

collinanderson:
Ok, the meta refactor has been merged. We could probably start working on getting compatible with 1.8 again.

phoebebright:
Now getting this error with 1.8a1

      File "/Users/phoebebr/.virtualenvs/gmd3/lib/python2.7/site-packages/taggit/managers.py", line 311, in post_through_setup
        self.related = ForeignObjectRel(cls, self.model, self)
    AttributeError: can't set attribute

with simplest model:

    class Question(models.Model):
        title    = models.CharField(max_length=200)
        tags = TaggableManager()

But ok with 1.7

RossLote:
I found this in django 1.8

    class RelatedField(Field):
        ...
        @property
        def related(self):
            warnings.warn(
                "Usage of field.related has been deprecated. Use field.rel instead.",
                RemovedInDjango20Warning, 2)
            return self.rel

So I'd think that this would fix it:
change this:

    self.related = RelatedObject(self.through, cls, self)

for this:

    self.rel = ForeignObjectRel(cls, self.model, self)

but then the TaggableManager rel attribute will have to be renamed

Any thoughts?

RossLote:
This is quite frustrating as this is one of two apps that will now be broke in my site and they aren't straight forward fixes. This:

> Private attribute django.db.models.Field.related is deprecated in favor of Field.rel. The latter is an 
> instance of django.db.models.fields.related.ForeignObjectRel which replaces 
> django.db.models.related.RelatedObject

to me implies that it's a drop in replacement but it just doesn't seem to work that way.

kaedroho:
I've got all the tests working on Django 1.8 beta 2 in my pull request (#297). Is it possible that you could try this on your sites and let me know if you hit any issues?

@RossLote 
[...]

PR #297, opened 2015-03-15
Django 1.8 support

kaedroho:
Addresses #284 

This brings Taggit up to speed with Django 1.8 beta 2. There may be more work to do before the final release but this should get us most of the way there.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,12 +7,15 @@ env:
   - TOXENV=py27-1.5.x
   - TOXENV=py27-1.6.x
   - TOXENV=py27-1.7.x
+  - TOXENV=py27-1.8.x
   - TOXENV=py33-1.5.x
   - TOXENV=py33-1.6.x
   - TOXENV=py33-1.7.x
+  - TOXENV=py33-1.8.x
   - TOXENV=py34-1.5.x
   - TOXENV=py34-1.6.x
   - TOXENV=py34-1.7.x
+  - TOXENV=py34-1.8.x
 install:
   - pip install flake8 tox
 before_script:
--- a/taggit/managers.py
+++ b/taggit/managers.py
@@ -231,6 +231,12 @@ class _TaggableManager(models.Manager):
 
 
 class TaggableManager(RelatedField, Field):
+    # Field flags
+    many_to_many = True
+    many_to_one = False
+    one_to_many = False
+    one_to_one = False
+
     _related_name_counter = 0
 
     def __init__(self, verbose_name=_("Tags"),
--- a/taggit/managers.py
+++ b/taggit/managers.py
@@ -243,12 +243,22 @@ class TaggableManager(RelatedField, Field):
                  help_text=_("A comma-separated list of tags."),
                  through=None, blank=False, related_name=None, to=None,
                  manager=_TaggableManager):
-        Field.__init__(self, verbose_name=verbose_name, help_text=help_text,
-                       blank=blank, null=True, serialize=False)
+
         self.through = through or TaggedItem
-        self.rel = TaggableRel(self, related_name, self.through, to=to)
         self.swappable = False
         self.manager = manager
+
+        rel = TaggableRel(self, related_name, self.through, to=to)
+
+        Field.__init__(
+            self,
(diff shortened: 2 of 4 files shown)


kaedroho:
All tests are now passing. Travis reports an error with Django 1.8/Python 2.7. This appears to be an issue with installing Django and I've sucessfully run the tests locally on Python 2.7/Django 1.8.

collinanderson:
Interesting. The 1.8 issue might be this https://github.com/django/django/commit/bd059e3f8c6311dcaf8afe5e29ef373f7f84cf26#commitcomment-10220493

Maybe switching from `master.zip` to `master.tar.gz` might fix it?

collinanderson:
We could also change the tag_model property to use `get_field()` on 1.8 so it doesn't raise a deprecation warning about `get_field_by_name`.

collinanderson:
I started switching the `get_field_by_name`'s but there's too many :). Probably out of scope for this PR.

collinanderson:
In any case, I tested this, and it at least works on _my_ site :)

taggit/managers.py
```diff
@@ -309,13 +339,18 @@ def __lt__(self, other):
         return False
 
     def post_through_setup(self, cls):
-        self.related = RelatedObject(cls, self.model, self)
+        if RelatedObject is not None:  # Django < 1.8
```
dtiesling: I'm able to import `RelatedObject` in django 1.8b2 so this logic is no longer working. Could you use django.VERSION for this instead?

kaedroho: related.py is gone in the latest code: https://github.com/django/django/tree/stable/1.8.x/django/db/models

Maybe it's a ``.pyc`` file hanging around, could you try reinstalling Django?

collinanderson: Interesting. If @dtiesling is running into that problem, I bet others will too. I wonder if we should use the VERSION method just to be safe.

.travis.yml
```diff
@@ -1,18 +1,26 @@
...
+   - env: TOXENV=py33-1.7.x
+   - env: TOXENV=py33-1.8.x
+   - env: TOXENV=py34-1.5.x
+   - env: TOXENV=py34-1.6.x
+   - env: TOXENV=py34-1.7.x
+   - env: TOXENV=py34-1.8.x
+  allow_failures:
+   - env: TOXENV=py27-1.8.x
+   - env: TOXENV=py33-1.8.x
+   - env: TOXENV=py34-1.8.x
```
collinanderson: Do you want to remove or revert the allow_failures commit?

kaedroho: Yes, remove it. Thanks

collinanderson:
Also, if you didn't notice, it looks like you need to bring the .tar.gz change back.

kaedroho:
Oops, thanks for spotting!

taggit/managers.py
dtiesling: thanks @kaedroho, you were right it was leftover files from not doing a clean install of django. I agree with @collinanderson I'm sure I won't be the only person to not think to clear out the old django before upgrading.

kaedroho: Fixed in 16f9f4a

collinanderson:
This looks good to me, at least :)

frewsxcv:
I'll merge this as soon as 1.8 final is out and i'll push a new taggit release.

chhantyal:
As 1.8 is released (https://www.djangoproject.com/weblog/2015/apr/01/release-18-final/), can we please accept this PR and update PyPi?

frewsxcv:
Has anyone tested this with the final release of 1.8?  The new version came out when I was on a flight. I'll look at merging this tomorrow.

frewsxcv:
Reopened to rerun the travis tests

troygrosfield:
:+1:

frewsxcv:
Great work @kaedroho! Thanks! Here is cake

frewsxcv:
0.13.0 has been published

https://pypi.python.org/pypi/django-taggit/0.13.0

let me know if any issues arise

PR #302, opened 2015-03-21
Find existing slugs in a custom Tag model

calvinpy:
Find existing slugs using default manager and not Tag model manager.
https://django-taggit.readthedocs.org/en/latest/custom_tagging.html?highlight=MyCustomTag


kaedroho:
@calvinpy This looks good to me, please could you add a test?

frewsxcv:
This was fixed in https://github.com/alex/django-taggit/pull/334, thanks for your pull request though!

PR #264, opened 2014-09-16
Issue #186 tentative fix

vhf:

PR #305, opened 2015-04-03
Fixes #186 and #264 Unknown column app.object_id Django 1.6

### LINKED ISSUE #186: Can't get django-taggit 0.10+ to work with Django 1.6.1
My tagged models work fine with Django 1.5.5, but my project throws an OperationalError: 1054 "Unknown column 'my_table.object_id in ...' with Django 1.6. This because my object's table doesn't have an object_id field; This field does exist in the taggit_taggeditem table, but the traceback shows that taggit is generating SQL referring to `my_table`.`object_id`...
This seems similar to this (open) Django bug: https://code.djangoproject.com/ticket/3055 - any idea what might be causing it or how to fix it?

kminkov:
This version is similar to the one provided in #264 but I think it shouldn't break the tests.


devoidfury:
This patch fixed the issue on a live site I'm upgrading; working now. +1

kminkov:
@frewsxcv Do we need something more to see this committed. If I could help with something just tell me. This is really annoying issue and would be great to fix it.

frewsxcv:
How difficult would it be to add a regression test for this?

kminkov:
I think it shouldn't be very hard. I will try to do it.

kminkov:
@frewsxcv , here is the test please take a look if everything is ok.

taggit/managers.py
```diff
@@ -484,14 +484,14 @@ def _get_gfk_case_path_info(self, direct=False):
         pathinfos = []
         from_field = self.model._meta.pk
         opts = self.through._meta
-        object_id_field = _get_field(self.through, 'object_id')
+        linkfield = self.through._meta.get_field_by_name(self.m2m_reverse_field_name())[0]
```
frewsxcv: Doesn't this variable get overwritten on the next line?

kminkov:
Yes, my mistake I forgot to clear it out.

frewsxcv:
Thanks for your contribution @kminkov!

frewsxcv:
0.17.1 has been released that includes this change

kminkov:
Perfect, you are welcome, and thanks for the cupcake :)

ChrisBraddock24:
Thanks.  This fix was very helpful!

PR #313, opened 2015-06-11
Add 'django-taggit-helpers' to external apps list in docs

mfcovington:
``django-taggit-helpers`` ([GitHub](https://github.com/mfcovington/django-taggit-helpers)/[PyPI](https://pypi.python.org/pypi/django-taggit-helpers)) makes it easier to work with admin pages of models associated with ``django-taggit`` tags by adding some helper classes:

  - ``TaggitCounter``: Display (and sort by) number of Taggit tags associated with tagged items.
  - ``TaggitListFilter``: Filter records by Taggit tags for the current model only. 
  - ``TaggitStackedInline``: Add stacked inline for Taggit tags to admin.
  - ``TaggitTabularInline``: Add tabular inline for Taggit tags to admin.


frewsxcv:
Thanks @mfcovington !

PR #316, opened 2015-06-12
Adds django-taggit-labels to external apps list

bennylope:
The label widget does not expose the input field so that you can add new tags. The base label widget instead shows all available tags and lets you pick between them in Django's admin interface. It presumes that you are using a managed tag list.


frewsxcv:
Thanks @bennylope !

PR #283, opened 2014-12-24
ISSUE: #282 , PK name on iherited models is not always 'id'

martende:
PK name on iherited models is not always 'id'

PR #322, opened 2015-06-24
Patch django-taggit to fix #282

### LINKED ISSUE #282: Bad JOIN collumn on tagging Inerited models
I have some simple inherited model
```
class Page(models.Model):
....
class Game(Page):
....
    tags = TaggableManager()
```
i I try to make 
```
games = Game.objects.filter(tags__name__in = [tagname])
```
I receive
OperationalError: (1054, "Unknown column 'game.id' in 'on clause'")

russmo:
django-taggit has a bug (#282) where it assumes that the tagged model
has a field named id which is the primary key. This isn't the case in
our database as we use a more descriptive pk/id field name. The django-taggit API lookup features fail unless this is fixed.

This commit patches the code using the fix in:
#283 and stevenday/django-taggit

### DIFF
--- a/taggit/managers.py
+++ b/taggit/managers.py
@@ -494,7 +494,7 @@ class TaggableManager(RelatedField, Field):
         if reverse_join:
             return ((self.model._meta.pk.column, "object_id"),)
         else:
-            return ((self.model._meta.pk.column, "id"),)
+            return (("object_id", self.model._meta.pk.column),)
 
     def get_extra_restriction(self, where_class, alias, related_alias):
         extra_col = _get_field(self.through, 'content_type').column
--- a/taggit/managers.py
+++ b/taggit/managers.py
@@ -492,9 +492,9 @@ class TaggableManager(RelatedField, Field):
 
     def get_joining_columns(self, reverse_join=False):
         if reverse_join:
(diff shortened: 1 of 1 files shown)


taggit/managers.py
```diff
@@ -492,9 +492,9 @@ def get_reverse_path_info(self):
 
     def get_joining_columns(self, reverse_join=False):
         if reverse_join:
-            return (("id", "object_id"),)
+            return ((self.model._meta.pk.column, "object_id"),)
         else:
-            return (("object_id", "id"),)
+            return ((self.model._meta.pk.column, "id"),)
```
frewsxcv: should this be `return (("object_id", self.model._meta.pk.column),)`?

svleeuwen: no, the patch is correct. i'm having the same issue, this fixes it.

russmo:
Sorry about that mixup in the order.  My bad.  I've fixed it and the pull request awaits your approval.

frewsxcv:
Thanks @russmo !

frewsxcv:
0.16.1 has been released that includes this fix

https://pypi.python.org/pypi/django-taggit/0.16.1

PR #324, opened 2015-07-01

mfeliu:
I don't think it makes sense to have tags which differ only in case. It's not even possible on MySQL because it violates the
unique constraint on tag name:

This PR adds a new setting, `TAGGIT_CASE_INSENSITIVE`, which causes existing tags to be looked up case insensitively. New
tags are added with the supplied case, which will be used from then on. The default behaviour is the same as before without
this setting.


mfeliu:
sorry, I did a mistake. I was trying to solve the merge-conflicts from another pull request.

PR #325, opened 2015-07-01

mfeliu:
This commit was adapted from an unmerged pull request [#263](https://github.com/alex/django-taggit/pull/263) done by [qris](https://github.com/qrisa). The following text is his original commit message:

I don't think it makes sense to have tags which differ only in case. It's not even possible on MySQL because it violates the
unique constraint on tag name:

This PR adds a new setting, `TAGGIT_CASE_INSENSITIVE`, which causes existing tags to be looked up case insensitively. New
tags are added with the supplied case, which will be used from then on. The default behaviour is the same as before without
this setting.

### DIFF
--- a/docs/getting_started.txt
+++ b/docs/getting_started.txt
@@ -30,3 +30,10 @@ And then to any model you want tagging on do the following::
         # ... fields here
 
         tags = TaggableManager()
+
+.. note::
+
+    If you want ``django-taggit`` to be **CASE INSENSITIVE** when looking up existing tags, you'll have to set to ``True`` the TAGGIT_CASE_INSENSITIVE setting (by default ``False``)::
+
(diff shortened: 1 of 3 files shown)


frewsxcv:
@mfeliu Seems alright to me. Can you do me one favor before merging? Can you just mention this setting exist somewhere in the documentation?

mfeliu:
@frewsxcv, I have including a note in the ``Getting Started`` section of the documentation.

frewsxcv:
Great job @mfeliu, thanks!

frewsxcv:
Released 0.16.0 that includes this change. Thanks again :)

https://pypi.python.org/pypi/django-taggit/0.16.0

mfeliu:
Thank you!

PR #327, opened 2015-07-12
Use international ISO 8601 dates in the changelog

svetlyak40wt:
Also, I've added dates for versions >= 0.9.2 (taken from git tags).


frewsxcv:
Great, thank you!

PR #334, opened 2015-08-06
fix #333 - TagBase._default_manager for existant tag check

### LINKED ISSUE #333: django.db.utils.IntegrityError with custom TagBase classes
As mentioned in https://github.com/alex/django-taggit/commit/b32c6cbeb37d40d00d687948ccc4854205b59b7c#commitcomment-12569645, the hard-coding of `Tag.objects` during the existence check of `TagBase.save()` means that other child classes of `TagBase` will not find existant duplicate tags while saving.

In apps with `TagBase` child classes, this can cause `django.db.utils.IntegrityError`s.

groovecoder:


frewsxcv:
Looks good. I'll merge this later today and ship a new version when I get home. How difficult would it be to create a regression test?

groovecoder:
I tried for a bit, but I'm not comfortable enough with the code-base to write it. :cry: I couldn't tell how to assert the behavior **inside** `TagBase.save()` was going thru the right flow.

frewsxcv:
That's alright, I understand. In case you missed it, flake8 is not happy about the style, stangely

groovecoder:
tumblr_m8x5o6nzc71rdkg05o1_500

taggit/models.py
```diff
@@ -68,8 +68,11 @@ def save(self, *args, **kwargs):
             except IntegrityError:
                 pass
             # Now try to find existing slugs with similar names
-            slugs = set(Tag.objects.filter(slug__startswith=self.slug)
-                                   .values_list('slug', flat=True))
+            slugs = set(
+                self._default_manager
```
jezdez: If I may suggest `_default_manager` should be accessed with the class not the instance as demonstrated in the majority of cases in Django's code base. If I understand correctly it has something to do how the default manager adds itself to the model class during import time and not the instance on creation time which in theory could be prevented if a Tag subclass were to fiddle with its class attributes during class or instance creation.

groovecoder: Can you link to one of those? I'm not as familiar with django internals, and most of the ones I see with a simple search show `model._default_manager`.

frewsxcv:
Thanks for the PR @groovecoder !

taggit/models.py
frewsxcv: If a decision here ever gets established, feel free to open another PR with the fix

frewsxcv:
Released 0.16.3 that includes this change. Thanks again!

https://pypi.python.org/pypi/django-taggit/0.16.3

taggit/models.py
jezdez: Ough I missed @groovecoder's comment. I'm away from the keyboard right now so can't link to all the cases in django's source code easily, but the `model` he refers to is a model class, not an instance of a model class. I'm not sure what else to say other than that if you want to use a private API like this (as it makes sense) I'd strongly suggest to follow the precedence to reduce the chance for surprises.

PR #337, opened 2015-08-18
Added ability to configure tag state

eugena:
I propose to add state field to tag. It seems to me very uncomfortable to create extra table for ability to configure tag state. Moreover, in that case, the query map is getting worse.

Also we can use custom setting of states by adding tuple TAG_STATES to Django settings.


auvipy:
This branch has conflicts that must be resolved

PR #329, opened 2015-07-14
Some early Django 1.9 compatibility fixes

kaedroho:
All tests pass when running against the latest Django master branch https://github.com/django/django/commits/8f8c54f70bfa3aa8e311514297f1eeded2c32593

Django 1.9 compatibility will probably break some more before the release, but I think this branch might be useful for people who want to do some early testing.

PR #343, opened 2015-09-24
Django 1.9 support for TaggableRel

nealtodd:
In the new alpha 1a release of Django 1.9 `TaggableRel.to` is replaced by `TaggableRel.model` (on the underlying `ForeignObjectRel` superclass.

This PR address that in a way that I think is compatible with < 1.9 (`to` as an attribute), 1.9 (`to` as a property) and > 1.9 (`to` removed). It's an example, you may want to implement it in a different way, so I'm not expecting it to be accepted necessarily.

The existing tests related to it pass. (Other tests fail with 1.9 alpha, but I've got a separate PR for that.)

### DIFF
--- a/taggit/managers.py
+++ b/taggit/managers.py
@@ -21,6 +21,16 @@ if VERSION < (1, 8):
 else:
     RelatedObject = None
 
+if VERSION < (1, 9):
+    # ForeignObjectRel.to is a property in Django 1.9 and removed in 2.0.
+    # It is replaced by ForeignObjectRel.model.
+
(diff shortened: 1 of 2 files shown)


nealtodd:
Just realised, some overlap with https://github.com/alex/django-taggit/pull/329. Sorry!

frewsxcv:
Closing in favor of https://github.com/alex/django-taggit/pull/352. Thanks for your PR though!

frewsxcv:
Just released 0.17.2 that has Django 1.9 fixes

https://pypi.python.org/pypi/django-taggit/0.17.2

Issue #349, opened 2015-10-19
Race condition with get_or_create on TaggedItem


mythmon:
When adding tags to items, the `TaggedItem` model is used for the intermediate table. In particular, `get_or_create` is used. There are no unique restrictions on this table, so [`get_or_create` is vulnerable to race conditions](https://docs.djangoproject.com/en/dev/ref/models/querysets/#get-or-create) if two threads/processes/servers try and add a tag to one item at the same time.

To fix this, I think it would work to add a unique_together contrainst on `(tag, content_type, object_id)` on the TaggedItem model. That way, the second creation will not happen (and `get_or_create` will recover correctly).

frewsxcv:
Thanks for pointing this out! If you create a pull request for this, I'd be happy to merge it

mythmon:
I filed a partial PR for this, but I'm having some trouble with migrations. I didn't see any examples or documentation about migrations. Can you provide some help in that PR?

PR #358, opened 2015-11-18
Add unique together constraint for TaggedItem.

mythmon:
This is supposed to fix the issue in #349. It isn't complete yet, but I've run into some problems.

* [x] Add unique together constraint to model.
* [x] Add Django migration
* [x] Add basic test
* [ ] Add South Migration
* [ ] Get subclassed tests to pass

I'm not sure how to go about about those last two. The Django migration I wrote by hand, but I'm familiar enough with South to not try and write a South migration by hand. I didn't see any obvious way to generate migrations, and the docs didn't have a "hacking on Taggit" section.

Any advice on how to finish this up?

### DIFF
--- a/taggit/migrations/0003_taggeditem_unique_together.py
+++ b/taggit/migrations/0003_taggeditem_unique_together.py
@@ -1,0 +1,18 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
+from django.db import models, migrations
+
+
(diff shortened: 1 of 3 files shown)


auvipy:
why there should be south migrations?

mythmon:
Django Taggit has other South migrations. It is probably good to keep maintaining those until they are removed entirely.

auvipy:
south migrations is removed

PR #359, opened 2015-11-20
Allows custom Through Model with GenericForeignKey

PierreF:
This pull request makes possible to use Django-taggit on any models with non-integer primary-key without the need to define one Through Model per tagged model (e.g. using GenericForeignKey not direct ForeignKey).

I moved the common code from GenericTaggedItemBase to a new CommonGenericTaggedItemBase class.

User may then inherit from CommonGenericTaggedItemBase to define a Through Model for custom PK while still using GenericForeignKey.

I've updated the doc to include example of a non-integer primary key and added test.

I've also included a GenericUUIDTaggedItemBase which is the same as GenericTaggedItemBase but for UUID primary-key (since it's probably the most common use case). But since it rely on Django 1.8, it's only defined for Django 1.8+

### DIFF
--- a/docs/custom_tagging.txt
+++ b/docs/custom_tagging.txt
@@ -10,6 +10,22 @@ want to store additional data about a tag, such as whether it is official.  In
 these cases ``django-taggit`` makes it easy to substitute your own through
 model, or ``Tag`` model.
 
+To change the behavior there are a number of classes you can subclass to obtain
+different behavior:
+
+=============================== =======================================================================
+Class name                      Behavior
+=============================== =======================================================================
+``TaggedItemBase``              Allows custom ``ForeignKeys`` to models.
+``GenericTaggedItemBase``       Allows custom ``Tag`` models. Tagged models use an integer primary key.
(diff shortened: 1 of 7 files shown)


frewsxcv:
Is it possible to make this backwards compatible? Right now it looks like we're renaming `GenericTaggedItemBase` to `CommonGenericTaggedItemBase` which could cause issues for people using it. Not a big deal if we have to do a new release with that breaking change, but if it's avoidable, that'd be ideal

PierreF:
``GenericTaggedItemBase`` still exists. I just moved most of the field & code to ``CommonGenericTaggedItemBase`` and ``GenericTaggedItemBase``. So using ``GenericTaggedItemBase`` works as it always did.

This is tested, since ``TaggedItem`` still use ``GenericTaggedItemBase`` and test suite pass.

PierreF:
Oh, and the previous `̀`GenericTaggedItemBase`` was already for tagged models that use an integer primary key. I just made that more clear in the doc.

frewsxcv:
Ah, that makes sense. Thanks for you contribution! :sparkles:

frewsxcv:
Just released version 0.17.4 that includes this change. Thanks again!

PR #361, opened 2015-11-24
fix #360 TaggableManager self.opts for Django 1.9

### LINKED ISSUE #360: Django commit c049c43 landing in 1.9 breaks TaggableManager
The https://github.com/django/django/commit/c049c43e28a3e46eaed13bb8e018b221250a0145 commit on Django 1.9.x stable (backported from master) and part of 1.9rc1 breaks TaggableManager during system checks, preventing an application from starting up:

```
Performing system checks...

Unhandled exception in thread started by <function wrapper at 0xb569579c>
Traceback (most recent call last):
  File "/django/utils/autoreload.py", line 226, in wrapper
    fn(*args, **kwargs)
  File "/django/core/management/commands/runserver.py", line 116, in inner_run
    self.check(display_num_errors=True)
  File "/django/core/management/base.py", line 426, in check
    include_deployment_checks=include_deployment_checks,
  File "/django/core/checks/registry.py", line 75, in run_checks
    new_errors = check(app_configs=app_configs)
  File "/django/core/checks/model_checks.py", line 28, in check_all_models
    errors.extend(model.check(**kwargs))
  File "/django/db/models/base.py", line 1170, in check
    errors.extend(cls._check_fields(**kwargs))
  File "/django/db/models/base.py", line 1249, in _check_fields
    errors.extend(field.check(from_model=cls, **kwargs))
  File "/django/db/models/fields/related.py", line 118, in check
    errors.extend(self._check_relation_model_exists())
  File "/django/db/models/fields/related.py", line 152, in _check_relation_model_exists
    rel_is_missing = self.remote_field.model not in self.opts.apps.get_models()
AttributeError: 'TaggableManager' object has no attribute 'opts'
```

I haven't figured out yet where `self.opts` gets set on a `Field` or `RelatedField` class (which `TaggableManager` subclasses) during app config. ~~I presume that because `TaggableManager` is slightly unusual in being a `django.db.models.Manager` as well as subclassing `RelatedField` and `Field` it might not be getting `self.opts` set in the way a model field would.~~ (edit: misread code, it's not a Manager, it's a field using _TaggableManager)

(Aside: why does `TaggableManager` subclass `RelatedField` and `Field` when `RelatedField` itself subclasses  `Field`?)

I'll do some more digging but I thought I'd report the issue now ahead of the scheduled Django 1.9 release in a few days (1st Dec).

collinanderson:


frewsxcv:
Is a regression test necessary or is it `master` currently failing?

collinanderson:
Looks like develop on 1.9 is _not_ failing currently, so, yes, I guess we need some sort of test. Is there anyway to simply run django's checks when testing?

collinanderson:
Like if we added `execute_from_command_line(sys.argv[:1] + ['check'])` to `runtests.py`, then it would show the failure.

frewsxcv:
Would it be possible to add it as part of the tox config? Would that be appropriate?

collinanderson:
You need the settings from runtests, but currently runtests doesn't handle anything other than the "test" command.

collinanderson:
I pushed a testcase that runs the django checks. This fails without self.opts.

frewsxcv:
That test case seems to cause issues on Django 1.4 and 1.5

collinanderson:
Ok, I fixed that except now I'm getting an issue with `ROOT_URLCONF` not being set. I think I need to stop for today. Feel free (anyone) to make an alternative PR.

nealtodd:
I think the only settings that are used in the tests are those defined `runtests.py`. A `ROOT_URLCONF` settings can't just be set in there though as there aren't any real urlpatterns for it to point to. Checks probably expect other settings too so adding settings to `runtests.py` is a bit of a non-starter.

However, for the purpose of this test we just want to see if there's something completely breaking startup, so we just need an individual check that doesn't care about settings. The `check` management command has a `tag` argument to specify which checks to run.

The `urls` check is the one that breaks because of `ROOT_URLCONF`. The `models` check is the one that detects the `AttributeError` that this PR fixes. So, a change to the test that works for me across the Django versions is:

```
    def test_django_checks(self):
        if django.VERSION >= (1, 6):
                call_command('check', tag=['models'])
        else:
                call_command('validate')
```

It's possible that other checks might detect breaking things in the future, which this test wouldn't detect, although the `models` check is probably the one most likely to break due to the nature of this package. However, short of putting all checks other than `urls` in the `tag` list (and updating it when new checks are added), this is at least better than no test at all.

There is a slight ugliness in that ``System check identified no issues (0 silenced).`` gets dumped into the test output. Specifying a verbosity of 0 doesn't supress it. The stdout argument of ``call_command`` could be used to dump it to /dev/null but that probably has issues on a Windows platform.

Issue #383, opened 2016-03-19
Switch to compiler.quote_name_unless_alias


orokusaki:
Running:

```
Spam.objects.aggregate(models.Count('tags'))
```

Produces this warning:

> [taggit/managers.py:87](https://github.com/alex/django-taggit/blob/develop/taggit/managers.py#L87): RemovedInDjango110Warning: Calling a SQLCompiler directly is deprecated. Call compiler.quote_name_unless_alias instead.

This isn't an issue yet, but I wanted to log it here, because it will be an issue with Django 1.10.

collinanderson:
See #380

orokusaki:
Ah. Thanks, Colin. Somehow, I missed that issue when I did an issue search.

jdufresne:
Looks like this has already been fixed. Thanks for reporting. If you continue to see an issue, please add additional information.

PR #380, opened 2016-03-05
Silence RemovedInDjango110Warning

ryokamiya:
Silence RemovedInDjango110Warning for the renamed/moved package.

https://docs.djangoproject.com/en/1.9/releases/1.8/#qn-replaced-by-compiler

PR #385, opened 2016-03-24
Fix all deprecation warnings

timheap:
I've fixed all deprecation warnings from all versions of Django supported by taggit, while maintaining backwards compatibility with all currently supported versions, and with all tests passing.

This involved adding some helpers for the recently renamed `field.rel.to`/`field.remote_field.model`, and changing the use of `add_lazy_relation` to `lazy_related_operation` on Django 1.9.

SQL names are quoted using the correct function now, fixing #380 and #383.

`on_delete=models.CASCADE` has been added to all ForeignKeys where it was missing, including in tests and migrations. This is backwards compatible with current code and projects, as `on_delete=models.CASCADE` was the assumed default previously.

Warnings due to different versions of unittest bundled through different version of Django have been sorted out, as has a warning from Django for Python 3.3, 3.4.

Finally, DeprecationWarnings are now show in test output. As there are no more deprecation warnings, the test output is still useable as it is before, without being polluted with lots of DeprecationWarning noise.

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -26,7 +26,7 @@ from .models import (Article, Child, CustomManager, CustomPKFood,
 from taggit.managers import _model_name, _TaggableManager, TaggableManager
 from taggit.models import Tag, TaggedItem
 
-from taggit.utils import edit_string_for_tags, parse_tags
+from taggit.utils import edit_string_for_tags, parse_tags, _remote_field, _related_model
 
 try:
     from unittest import skipIf, skipUnless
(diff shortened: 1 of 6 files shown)


tests/tests.py
```diff
@@ -56,7 +57,22 @@ def _get_form_str(self, form_str):
         return form_str
 
     def assert_form_renders(self, form, html):
-        self.assertHTMLEqual(str(form), self._get_form_str(html))
+        # Django causes a DeprecationWarning on Python 3.3, 3.4
+        if (3, 3) >= sys.version_info > (3, 5):
```
frewsxcv: Won't this always evaluate to `False`?

timheap: Good catch. Which means if the DeprecationWarnings are not appearing, that particular issue must have gone away... I'll look in to it.

timheap:
I fixed the backwards conditional that you spotted, properly silencing the DeprecationWarning from Django in those versions of Python.

frewsxcv:
Sorry for the delay! Thanks a bunch, your contributions are greatly appreciated :sparkles:

PR #394, opened 2016-04-19
Added info about applying migrations to README

arogachev:


arogachev:
Missed link with more info.

PR #396, opened 2016-04-27
Support for unicode tags and slugs

nilnvoid:


frewsxcv:
@nilnvoid Any chance you could a test or two for this? I'm also wondering this should sit behind a `settings.py` flag...

nilnvoid:
@frewsxcv You mean to write a test for this? I will look into this. And did not get what you mean by "this should sit behind a settings.py flag..."

frewsxcv:
Ignore my comment about the settings.py flag. Ping me if you ever get around to writing a test or two for this, thanks!

PR #397, opened 2016-04-29
Fix incorrect args in custom tagging example

prydie:
Original code example declares `tag_string` as the argument for `comma_joiner`, however, the body of `comma_joiner` expects an iterable of `Tag` objects named `tags`.


frewsxcv:
Thanks!

PR #402, opened 2016-05-11
Added spanish and turkish translation

SalahAdDin:
Added localization files for Spanish and Turkish language.


frewsxcv:
Thanks!

PR #408, opened 2016-05-19
Code and tooling improvements

prydie:
Small PR to start with. The main points are:

 - `isort` should now be run by travis
 - test results should be upload to codecov from travis
 - `E302` inconsistencies fixed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,8 @@
+.PHONY: develop test
+
+develop:
+	pip install -q -r requirements/test.txt
+	pip install -q -e .
+
+test: develop
+	DJANGO_SETTINGS_MODULE=tests.settings py.test tests/tests.py --cov=taggit
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,15 @@
+[run]
+branch = True
+source = taggit
+
+[report]
+exclude_lines =
+    if self.debug:
+    pragma: no cover
+    raise NotImplementedError
+    if __name__ == .__main__.:
+ignore_errors = True
+omit =
+    tests/*
+    taggit/migrations/*
+    taggit/south_migrations/*
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,6 @@ django_taggit.egg-info/
 dist/
 _build/
 .tox/
+.cache/
+.coverage
+.eggs/
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,7 +49,7 @@ script:
 
 after_success:
   - pip install codecov
-  - codecov -e TOX_ENV TRAVIS_OS_NAME
+  - codecov -e TOXENV CI TRAVIS_BUILD_ID TRAVIS TRAVIS_BRANCH TRAVIS_JOB_NUMBER TRAVIS_PULL_REQUEST TRAVIS_JOB_ID TRAVIS_REPO_SLUG TRAVIS_COMMIT
 
 notifications:
   email: false
--- a/.travis.yml
+++ b/.travis.yml
@@ -37,13 +37,11 @@ matrix:
     - env: TOXENV=py35-1.9.x
       python: 3.5
 install:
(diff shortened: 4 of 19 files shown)


frewsxcv:
Looks great! Though we need to use a tool to generate code coverage. codecov just looks for the `.coverage` file and uploads it to the servers. Since it doesn't exist, coverage isn't working.

```
  _____          _
 / ____|        | |
| |     ___   __| | ___  ___ _____   __
| |    / _ \ / _` |/ _ \/ __/ _ \ \ / /
| |___| (_) | (_| |  __/ (_| (_) \ V /
 \_____\___/ \__,_|\___|\___\___/ \_/
                                0db33a1
==> Travis CI detected.
(root) .
==> Running gcov (disable via -X gcov)
  -> gcov in .
==> Searching for coverage reports
    Found 1 reports.
**> Python coverage not found
**> No coverage report found.
```

frewsxcv:
In other words, http://coverage.readthedocs.io/en/coverage-4.0.3/ will probably need to be incorporated.

prydie:
I'm working on it at the moment actually. The issue I'm coming up against now is that now I'm [integrating it with tox](https://github.com/codecov/example-python#using-tox) it's also being run when you run tox locally. This only causes the below message to be displayed for each testenv but it's by no means ideal.

```
py27-1.9.x runtests: commands[1] | codecov -e TOXENV

      _____          _
     / ____|        | |
    | |     ___   __| | ___  ___ _____   __
    | |    / _ \ / _  |/ _ \/ __/ _ \ \ / /
    | |___| (_) | (_| |  __/ (_| (_) \ V /
     \_____\___/ \____|\___|\___\___/ \_/
                                    v2.0.3

==> Detecting CI provider
  -> Got branch from git/hg
  -> Got sha from git/hg
==> Preparing upload
Error: Missing repository upload token

Tip: See all example repositories: https://github.com/codecov?query=example
Support channels:
  Email:   hello@codecov.io
  IRC:     #codecov
  Gitter:  https://gitter.im/codecov/support
  Twitter: @codecov
```

frewsxcv:
I'm fine putting the codecov logic in tox, though I think it would also be okay to leave it in the Travis CI file since we only want it to upload to their servers in the CI environment.

prydie:
@frewsxcv It should now be ready for merging. I have added the coverage and build buttons to the `README.rst`, vastly simplified the `tox.ini` and coverage.io is *finally* working.

I'm happy to rebase the branch and create a new pull request if you'd prefer to keep the history clean (or indeed "squash and merge" may be appropriate).

Makefile
```diff
@@ -0,0 +1,8 @@
+.PHONY: develop test
+
```
frewsxcv: I'm not really a fan of Makefiles personally. Is it possible we could just reuse the existing `runtests.py` file? Otherwise, it's not that big of a deal I guess

frewsxcv:
This is great! Just left one comment, but then this should be good to go! 🎊

Makefile
prydie: I don't think `py.test` would work with the existing `runtests.py` code (but I may be wrong). `Makefile`s are used across a number of popular Django projects in this way (see [wagtail](https://github.com/torchbox/wagtail/blob/master/Makefile), [django-model-utils](https://github.com/carljm/django-model-utils/blob/master/Makefile), [django-activity-stream](https://github.com/justquick/django-activity-stream/blob/master/Makefile) etc.). What about them do you dislike?

frewsxcv: In your particular examples (and the one in this PR) I don't have many complaints since it's so simple, but I think the syntax is unintuitive at times and results in a giant mess of spaghetti (:spaghetti: 💻 ). Anyways, what you have here seems simple enough so no reason to hold this PR up.

prydie:
> Thanks for your contribution!

My pleasure :smile:

PR #409, opened 2016-05-20
Implementation of m2m_changed signal sending

prydie:
I have implemented the firing of the [`m2m_changed`](http://devdocs.io/django~1.9/ref/signals#m2m-changed) signal. This signal is normally generated by the [`ManyRelatedManger`](https://github.com/django/django/blob/67d984413c9540074e4fe6aa033081a35cf192bc/django/db/models/fields/related.py#L1074) class, however, as `TaggableManager` is not a subclass of [`ManyToManyField`](http://devdocs.io/django~1.9/ref/models/fields#django.db.models.ManyToManyField) this functionality is missing in taggit.

There are a number of areas on which I'd like a second opinion; please see line comments.


taggit/managers.py
```diff
@@ -203,9 +206,31 @@ def add(self, *tags):
         for new_tag in tags_to_create:
             tag_objs.add(self.through.tag_model().objects.create(name=new_tag))
 
+        new_ids = set(t.pk for t in tag_objs)
+
+        # NOTE: can we hardcode 'tag_id' here or should the column name be got
```
prydie: The `TaggableManagerOfficialTestCase` passes and the [custom tag docs](http://django-taggit.readthedocs.io/en/latest/custom_tagging.html#custom-tag) suggest so but a second opinion would be good.

taggit/managers.py
```diff
@@ -203,9 +206,31 @@ def add(self, *tags):
...
+        # dynamically from somewhere?
+        vals = (self.through._default_manager.using(db)
+                .values_list('tag_id', flat=True)
+                .filter(**self._lookup_kwargs()))
+
+        new_ids = list(new_ids - set(vals))
+
+        signals.m2m_changed.send(
+            sender=self.through, action="pre_add",
+            instance=self.instance, reverse=False,
```
prydie: I think it is safe to hard code `reverse=False` here as as I believe `_TaggableManager` is always used as a forward many-to-many descriptor, however, confirmation of this would be great.

taggit/managers.py
```diff
@@ -221,13 +246,47 @@ def set(self, *tags):
...
     @require_instance_manager
     def remove(self, *tags):
-        self.through.objects.filter(**self._lookup_kwargs()).filter(
-            tag__name__in=tags).delete()
+        if not tags:
+            return
+
+        db = router.db_for_write(self.through, instance=self.instance)
+
+        qs = (self.through._default_manager.using(db)
```
prydie: This introduces an additional SQL query, however, core's `ManyRelatedManager` also performs this query so I don't think it's unwarranted.

taggit/managers.py
```diff
@@ -203,9 +206,31 @@ def add(self, *tags):
         for new_tag in tags_to_create:
             tag_objs.add(self.through.tag_model().objects.create(name=new_tag))
 
+        new_ids = set(t.pk for t in tag_objs)
+
+        # NOTE: can we hardcode 'tag_id' here or should the column name be got
+        # dynamically from somewhere?
+        vals = (self.through._default_manager.using(db)
```
prydie: This adds an additional SQL query, however, as the method already produces `O(n)` SQL queries it's rather semantic. There is potentially room for using `bulk_create` to achieve `O(1)`, however, that is outside the scope of this PR.

taggit/managers.py
```diff
@@ -203,9 +206,31 @@ def add(self, *tags):
...
+                .values_list('tag_id', flat=True)
+                .filter(**self._lookup_kwargs()))
+
+        new_ids = set(new_ids - set(vals))
+
+        signals.m2m_changed.send(
+            sender=self.through, action="pre_add",
+            instance=self.instance, reverse=False,
```
prydie: I think it is safe to hard code `reverse=False` here as as I believe `_TaggableManager` is always used as a forward many-to-many descriptor, however, confirmation of this would be great.

prydie:
@frewsxcv I asked @apollo13 to take a look in IRC and he suggested 8571c3c to ensure that the `using` argument passed to the signal is always correct. Other than that I think this is starting to look ready for merging. Thoughts?

prydie:
Having said that I've found another problem that requires an API breaking change to fix in full.

There is an inconsistency between taggit's [`_TaggableManager.set()`](http://django-taggit.readthedocs.io/en/latest/api.html#TaggableManager.set) API and Django's [`ManyRelatedManager.set()`](https://github.com/django/django/blob/ed0ff913c648b16c4471fc9a9441d1ee48cb5420/django/db/models/fields/related_descriptors.py#L907). This results in the `m2m_changed` signals being incorrect when you update the fields of an existing object via the form field (`action='{pre,post}_clear'` rather than `action='{pre,post}_remove'`).

#### Expected behavior

Any existing objects in the `through` table that are not present in the arguments to `_TaggableManager.set()` are removed (using `_TaggableManager.remove()`) and any new objects are added with `_TaggableManager.add()`.

#### Actual behavior

> `_TaggableManager.set()`
> Removes **all** the current tags _(by calling `_TaggableManager.clear()`)_ and then adds the specified tags to the object _(by calling `_TaggableManager.add()`)_.

_(From the docs, emphasis and brackets mine)_

What makes this worse is that the method signature has diverged from core's (core expects `set(objs, **kwargs)` whereas we expect `set(*tags)`).

We can either:

 1. Change the API to be in line with core's and change the behavior so that internally we use `_TaggableManager.remove()`
 2. Keep the existing API but change the behavior so that internally we use `_TaggableManager.remove()`

I will go ahead and work on changing the behavior whilst maintaining the current API for now. I suggest if we're going to do an API breaking change we identify all of the inconsistencies and fix them in a single release.

taggit/managers.py
frewsxcv: Sounds good to me 👍

taggit/managers.py
frewsxcv: Sounds reasonable

taggit/managers.py
```diff
@@ -185,26 +188,52 @@ def add(self, *tags):
...
+
+        new_ids = set(t.pk for t in tag_objs)
+
+        # NOTE: can we hardcode 'tag_id' here or should the column name be got
+        # dynamically from somewhere?
+        vals = (self.through._default_manager.using(db)
+                .values_list('tag_id', flat=True)
+                .filter(**self._lookup_kwargs()))
+
+        new_ids = set(new_ids - set(vals))
```
frewsxcv: This could also be `set(new_ids) - set(vals)`

taggit/managers.py
```diff
@@ -215,18 +258,82 @@ def slugs(self):
         return self.get_queryset().values_list('slug', flat=True)
 
     @require_instance_manager
-    def set(self, *tags):
-        self.clear()
-        self.add(*tags)
+    def set(self, *tags, **kwargs):
```
prydie: Implementation intended to match [`ManyRelatedManager.set()`](https://github.com/django/django/blob/ed0ff913c648b16c4471fc9a9441d1ee48cb5420/django/db/models/fields/related_descriptors.py#L907) as closely as possible without breaking the existing API.

taggit/managers.py
prydie: In fact, it should be as follows as `new_ids` is already a set.
 
```python
new_ids = new_ids - set(vals))
```

frewsxcv:
Everything looks good here to me! Have anything else to add or is it ready to be merged?

prydie:
> Everything looks good here to me! Have anything else to add or is it ready to be merged?

:tada: Great news. As far as I'm concerned it's ready to be merged. It's now passing integration tests for the Django app I'm developing that requires this functionality.

Are you in a position to do a PyPi release once it's merged?

frewsxcv:
Awesome work, thanks again for your contribution! 🎉 

> Are you in a position to do a PyPi release once it's merged?

Yep, will do this immediately after.

frewsxcv:
Version 0.19.0 has been published which includes these changes. Thanks again!

PR #410, opened 2016-05-23
add apps.py and translation file

9nix00:
thanks django-taggit, it saves my time!

this is my contribute. hope this is useful. please check it.


prydie:
CI is failing due to icorrectly ordered imports in `taggit/apps.py`. If you could fix this so that the rest of the test suite runs that would be great.

```
ERROR: /home/travis/build/alex/django-taggit/taggit/apps.py Imports are incorrectly sorted.
```

9nix00:
upgrade,please confirm it.

taggit/apps.py
```diff
@@ -0,0 +1,6 @@
+from django.apps import AppConfig as BaseConfig
+from django.utils.translation import ugettext_lazy as _
+
+class AppConfig(BaseConfig):
```
prydie: The standard around App config is to name the config class based on the module name. For example:

 ```python
 # rock_n_roll/apps.py

from django.apps import AppConfig

class RockNRollConfig(AppConfig):
    name = 'rock_n_roll'
    verbose_name = "Rock ’n’ roll"
```
From the [docs](https://docs.djangoproject.com/en/1.9/ref/applications/#for-application-authors).

prydie:
One suggestion code wise. Not being a Chinese speaker though I can't review the translation, unfortunately.

taggit/apps.py
frewsxcv: In other words, just rename this to `TaggitAppConfig` and this should be good to merge

frewsxcv:
Looks great, thanks!

PR #414, opened 2016-06-10
Custom slugify

PixxxeL:
In some cases slugify not working. Bacause I suggest injection of customer slugify through settings.

Additionaly, I'm translate name of application on Russian.


taggit/locale/ru/LC_MESSAGES/django.po
```diff
@@ -17,6 +17,9 @@ msgstr ""
 "Plural-Forms: nplurals=3; plural=n%10==1 && n%100!=11 ? 0 : n%10>=2 && n%10<=4 && (n%100<10 || n%100>=20) ? 1 : 2;\n"
 "X-Poedit-Language: Russian\n"
 
+msgid "Taggit"
+msgstr "Метки"
```
frewsxcv: Isn't this already specified on line 30 of this file?

PixxxeL: In line 30 specified `msgid "Tags"`, but here is `msgid "Taggit"` that define in [taggit/apps.py](https://github.com/PixxxeL/django-taggit/blob/custom-slugify/taggit/apps.py). And my check are show that this string is not translate. But after Im add `msgid "Taggit"` - all was fine.

prydie:
I'm of the opinion that our [current advice](http://django-taggit.readthedocs.io/en/latest/custom_tagging.html#TagBase.slugify) on slugify customization is preferable to peppering the Django settings file more developer selected "knobs" to twiddle.

Also, if we do go down the route of adding this as a Django setting then I would suggest `TAGGIT_SLUGIFY` would be a more idiomatic name. I'm not a fan of `METH` as a shorthand for `METHOD` and given that the setting would actually be defining a function (or a static method I guess) it doesn't make sense. Furthermore, we already have `TAGGIT_TAGS_FROM_STRING` and `TAGGIT_STRING_FROM_TAGS` that follow this naming convention.

Anyway, Just my 2 cents. Thanks for your contribution :+1:

PixxxeL:
I thought that the chosen method does not exclude the use of other methods, and particularly does not overhead. Often the developer is save time to customization and would like to set up and forget.

And about `TAGGIT_SLUGIFY` I'm agree. Long thought about name of setting, but have got the better of me Python's "Explicit is better than implicit." Maybe then this setting to rename if this pullrequest is arranges?

frewsxcv:
I just merged https://github.com/alex/django-taggit/pull/414/commits/fd65f0b26f0f3cd77a3263f5e7b6f3961cc5e0ee in https://github.com/alex/django-taggit/pull/416.

I'm hesitant to merge in https://github.com/alex/django-taggit/pull/414/commits/c1efc2654842ca2411707800d4ef34520b76199d. In general, I prefer subclassing over global variables. Concretely, this means subclassing `TagBase` and overriding `slugify` instead of specifying a new setting in `settings.py`.

PixxxeL:
If inherit `TagBase`, is it `Tag` model (that already exist and not abstract) not create table and then `CustomTag` (as example) is create? How it is possible?

frewsxcv:
I'm not sure I understand your question, but if you create a `CustomTag` model that is a subclass of `TagBase`, it *will* be a new database table `<model name>_customtag`. You can override the name of the table using the [`Meta.db_table` setting](https://docs.djangoproject.com/en/1.9/ref/models/options/#db-table).

example:

```python
class CustomTag(TagBase):
    class Meta:
        db_table = "tag"
```

frewsxcv:
A more thorough example can be found here: https://django-taggit.readthedocs.io/en/latest/custom_tagging.html#custom-tag

PixxxeL:
> You can override the name of the table
Fine! Thanks!

PR #386, opened 2016-03-31
Clean up

auvipy:
first pass.

removed django 1.4-1.6

PR #424, opened 2016-07-19
resolve related_field_name dynamically

nickhudkins:
We ran into a problem where multi-inheritance + lazy ForeignKey resolution resulted in a KeyError in `.similar_objects()`

This PR addresses the problem by doing the remote field name resolution dynamically. The result is that in certain cases, the object returned by `getattr(obj, actual_remote_field_name)` was an actual instance rather than the pk of the desired relation.

I have added a test case for this which is now passing with the commit in `taggit/managers.py`


frewsxcv:
As per [the Travis CI tests](https://travis-ci.org/alex/django-taggit/builds/145886315), this doesn't appear to be compatible with Django <1.9. Do you know of an easy way to do that?

nickhudkins:
@frewsxcv waiting on the build to finish, tested locally with 1.8. We'll see how it goes :)

frewsxcv:
I just realized that https://github.com/alex/django-taggit/pull/386 merged into the wrong branch. So don't worry about 1.4-1.6 failures right now as support for those versions of Django will be dropped shortly.

nickhudkins:
Alright, it looks like I just have some bad migrations in there now. I'll resolve later today as it currently is working for my RealWorld™ use 👍

frewsxcv:
Going to take care of https://github.com/alex/django-taggit/issues/425 before this pull request can move forward.

frewsxcv:
#425 has been addressed. When you get a chance, you should rebase this off master and see if the tests pass

murdav:
Any update on this pull request?
Thanks!

nemesisdesign:
@nickhudkins could you try to follow @frewsxcv's suggestion?

taggit/managers.py
```diff
@@ -365,8 +365,13 @@ def similar_objects(self):
             objs = rel_model._default_manager.filter(**{
                 "%s__in" % remote_field.field_name: [r["content_object"] for r in qs]
             })
+            actual_remote_field_name = remote_field.field_name
+            if VERSION > (1, 9):
```
jdufresne: taggit no long supports Django < 1.11. Could you update the PR to drop workarounds for older Djangos?

PR #429, opened 2016-07-28
Address list_display and fieldsets in admin docs

jcushman:
A couple of additions to the docs for the Django admin -- explain how to include tags in fieldsets (this seemed obvious in retrospect but took me a few minutes to figure out), and provide a sample ModelAdmin that efficiently includes a list of tags in list_display.


frewsxcv:
Thanks a bunch! 🎉

Issue #432, opened 2016-08-18
Test failures with Django 1.10


nijel:
When running testsuite on Django 1.10 several tests fail:

```
======================================================================
FAIL: test_form (tests.tests.TaggableFormCustomPKTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/nijel/work/debian/django-taggit/tests/tests.py", line 790, in test_form
    <tr><th><label for="id_tags">Tags:</label></th><td><input type="text" name="tags" value="green, red, yummy" id="id_tags" /><br />%(help_start)sA comma-separated list of tags.%(help_stop)s</td></tr>""")
  File "/home/nijel/work/debian/django-taggit/tests/tests.py", line 66, in assertFormRenders
    self.assertHTMLEqual(str(form), self._get_form_str(html))
  File "/home/nijel/work/debian/django-taggit/django/test/testcases.py", line 699, in assertHTMLEqual
    self.fail(self._formatMessage(msg, standardMsg))
AssertionError: <tr>
<th>
<label for="id_name">
Name:
</label>
</th><td>
<input id="id_name" max [truncated]... != <tr>
<th>
<label for="id_name">
Name:
</label>
</th><td>
<input id="id_name" max [truncated]...
  <tr>
  <th>
  <label for="id_name">
  Name:
  </label>
  </th><td>
- <input id="id_name" maxlength="50" name="name" required type="text" value="apple" />
?                                                ---------

+ <input id="id_name" maxlength="50" name="name" type="text" value="apple" />
  </td>
  </tr><tr>
  <th>
  <label for="id_tags">
  Tags:
  </label>
  </th><td>
- <input id="id_tags" name="tags" required type="text" value="green, red, yummy" /><br /><span class="helptext">
?                                 ---------

+ <input id="id_tags" name="tags" type="text" value="green, red, yummy" /><br /><span class="helptext">
  A comma-separated list of tags.
  </span>
  </td>
  </tr>

======================================================================
FAIL: test_form (tests.tests.TaggableFormDirectCustomPKTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/nijel/work/debian/django-taggit/tests/tests.py", line 790, in test_form
    <tr><th><label for="id_tags">Tags:</label></th><td><input type="text" name="tags" value="green, red, yummy" id="id_tags" /><br />%(help_start)sA comma-separated list of tags.%(help_stop)s</td></tr>""")
[...]

PR #433, opened 2016-08-22
Fix form tests on Django 1.10

jdufresne:
Django 1.10 now adds the "required" attribute to inputs. Handle this in
tests.

Refs #432


frewsxcv:
Thanks!

Issue #33, opened 2010-07-06
Autocomplete widget for the admin.


alex:
^

frewsxcv:
Related pull request: https://github.com/alex/django-taggit/pull/441

charlex:
@frewsxcv #441 doesn't implement autocomplete, though it leaves the door open for it. I didn't want to include it because it felt like a can of worms, with its own set of problems, like adding an extra query to load all the available tags. The plugin I'm using definitely does have autocomplete built in.

SalahAdDin:
What's about this?

rtpg:
I think this might be an interesting thing to tackle once we resolve the #707 issue. If we vendor it in we could use it to tackle this stuff as well

regisb:
At [Open edX](https://open.edx.org), we are using django-taggit-autosuggest which is no longer maintained and not compatible with Django 3.2. Would you be interested in vendoring django-taggit-autosuggest in django-taggit? If yes I can open a PR.

mihalt:
@rtpg so, what is the situation with autocomplete native support? Do you plan to introduce? It would be really helpful in admin panel.

mihalt:
@rtpg or maybe there are any widgets like m2m field for taggit where I can select from predefined earlier tags? I have some models where in their admin panels not very necessary to create new tags from the changeview.

PR #441, opened 2016-09-08
Easy-to-use jQuery Tag-It Widget

charlex:
From a developer's perspective I love django-taggit. However, I realize that from a user's POV, writing and managing tags can feel cumbersome as they have to remember to format their input using a combo of quotes, commas or spaces. This is why I wrote this PR. 

The intent of this code is to, by default, give users a simple widget that solves the mentioned problem and makes creating and deleting tags much easier. 

It accomplishes this goal by replacing the text input widget with an interactive input field powered by [jQuery Tag-It](http://aehlke.github.io/tag-it/). It looks like this:

screen shot 2016-09-08 at 2 40 50 pm

My hope is that this makes django-taggit an off-the-shelf complete solution for developers and that this plugin provides a framework that accommodates the planned roadmap.

I haven't written any unit tests for this, as I'm not sure how to test a javascript widget, but I'd be happy to take an attempt if someone has a good approach for it.


charlex:
I'm closing this till I can get tests working.

charlex:
Tests are working and my code changes are now less obtrusive than they were before.

charlex:
Alright, for some reason the tests are working great on my local environment, but on Travis they are failing. It looks like maybe my test settings that are set up to use the widget template are not working. 

Does anyone have an idea on how to get the new settings to be picked up properly?

taggit/utils.py
```diff
@@ -46,13 +46,7 @@ def _parse_tags(tagstring):
...
     tagstring = force_text(tagstring)
 
-    # Special case - if there are no commas or double quotes in the
-    # input, we don't *do* a recall... I mean, we know we only need to
-    # split on spaces.
-    if ',' not in tagstring and '"' not in tagstring:
-        words = list(set(split_strip(tagstring, ' ')))
-        words.sort()
-        return words
+    print tagstring
```
frewsxcv: This probably shouldn't be here, nor is it Python 3 compatible

charlex: Fixed. Thanks for catching that, it was an accident.

frewsxcv:
I did a quick glance through the code and nothing stuck out as wrong to me. Maybe a misconfiguration with https://docs.djangoproject.com/en/1.10/ref/templates/api/#django.template.loaders.app_directories.Loader ?

charlex:
@frewsxcv Thanks for the suggestion. I took a quick stab at trying to configure it correctly but didn't get it there just yet. I'm out of time right now to continue looking at it, but tomorrow  I should have some more time to do a deeper dive.

clintonb:
@charlex @frewsxcv why was this closed? Are there other efforts to add autocomplete?

clintonb:
Never mind. I see the comment on #33

> ...doesn't implement autocomplete, though it leaves the door open for it. I didn't want to include it because it felt like a can of worms, with its own set of problems, like adding an extra query to load all the available tags. The plugin I'm using definitely does have autocomplete built in.

PR #455, opened 2017-01-10
Add __gt__ and __lt__ overrides for TagBase.

adam-thomas:
Fixes an incompatibility with django-modelcluster and, by extension, Wagtail.

Currently untested, just committing the change to 'save' it.


adam-thomas:
Oops, created this by accident.  Please ignore. :)

PR #456, opened 2017-01-10
Support __gt__ and __lt__ ordering on Tags

adam-thomas:
This makes it possible to call `sort()` on a list of `Tag` objects in memory.


frewsxcv:
Thanks!

adam-thomas:
Thank you :)

PR #465, opened 2017-01-29
Drop support for Django 1.7 and remove all workarounds

jdufresne:
Django 1.7 is no longer supported and has not received security updates since
2015-12-01. This avoids workarounds for potentially insecure and outdated
Django.

See table of Django supported versions:

https://www.djangoproject.com/download/#supported-versions


frewsxcv:
Thanks!

PR #473, opened 2017-03-22
Update spanish translation

tonicabanas:
We are use to write "slug" in Spanish instead of "marquilla", that is a pretty strange word...


frewsxcv:
Thanks!

PR #482, opened 2017-04-19
Improved auto-slug in TagBase to support UUID pk

### LINKED ISSUE #478: Automatic slug doesn't work when using UUID primary keys
[The check contained in TagBase](https://github.com/alex/django-taggit/blob/825c4b4c402691be3b14e825d609982c4a3e0686/taggit/models.py#L38) doesn't work on custom tag models that set a UUID primary key, that's because the following check:

```python
if not self.pk and not self.slug:
    # ...
```

is never triggered since in a UUID based model `self.pk` is never `None`.

I'm preparing a pull request to address this issue.

nemesisdesign:
Replaced `not self.pk` with the more recent `self._state.adding`,
this allows to determine if the model instance is new also
when using different primary keys (eg: UUID fields)

Fixes #478


nemesisdesign:
@frewsxcv I fixed tests for older django versions (1.8 and 1.9). I think it's safe to merge since the `_state` attribute on models is available since django 1.8.

frewsxcv:
Thanks!

Issue #75, opened 2011-06-17
Getting wrong content_type_id when using Inheritance


meir90:
Hello,

I defined:

```
from django.db import models
from taggit.managers import TaggableManager

class Parent(models.Model):
  tags = TaggableManager()
class Child (models.Model):
   more_stuff = models.CharField(max_length=200, blank=True)
```

When I add a Child element using the admin panel (and using Inline I get the "TAGS" field that should be added to the related Parent element) then the content_type_id in in taggit_taggeditem is wrong and points to Child instead of Parent...

vst:
Is there any update/workaround for this?

nemesisdesign:
I suggest to open a pull request with a failing test, that will help maintainers to understand the issue.

rtpg:
The docs [have a thing on this now](https://django-taggit.readthedocs.io/en/latest/forms.html), you can also [provide a custom string parser now](https://django-taggit.readthedocs.io/en/latest/custom_tagging.html#using-a-custom-tag-string-parser). 

There might be an argument to change the defaults now that we have a configuration thing, but that's a bit of a separate thing

PR #483, opened 2017-04-27
Tests to Re-produce Inheritance Problem

vst:
In relation to #75, as per @nemesisdesign's suggestion.

I guess that the added test is pretty much self explanatory.


auvipy:
hey, could you rebase this? and do you have any solution in mind? if you are looking for takers of this, please tell. thanks for your efforts.

vst:
@auvipy Sorry, I have missed your comment. 

It is a 2+ years old topic, and I am no longer using inheritance as I did before. For the sake of sanity, I am closing the PR.

PR #487, opened 2017-05-07
Add missing tests for required primary key

atombrella:
I searched, and this seems like one possible way to test that's something has been saved or not.

### DIFF
--- a/tests/tests.py
+++ b/tests/tests.py
@@ -428,7 +428,12 @@ class TaggableManagerTestCase(BaseTaggingTestCase):
 
     def test_require_pk(self):
         food_instance = self.food_model()
-        self.assertRaises(ValueError, lambda: food_instance.tags.all())
+        msg = (
+            '%s objects need to have a primary key value before you can access '
+            'their tags.' % self.food_model().__class__.__name__
+        )
+        with self.assertRaisesMessage(ValueError, msg):
+            food_instance.tags.all()
 
     def test_delete_obj(self):
         apple = self.food_model.objects.create(name="apple")
@@ -671,8 +676,8 @@ class TaggableManagerDirectCustomPKTestCase(TaggableManagerTestCase):
     taggeditem_model = TaggedCustomPKFood
 
     def test_require_pk(self):
-        # TODO with a charfield pk, pk is never None, so taggit has no way to
-        # tell if the instance is saved or not
+        # With a CharField pk, pk is never None. So taggit has no way to tell
+        # if the instance is saved or not.
         pass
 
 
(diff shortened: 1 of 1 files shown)


nemesisdesign:
@atombrella thank you for helping out. I do not understand the purpose of the test, the comment in the test class is confusing.

How is the missing primary key if using a char field as primary key related to django-taggit?
What's the expected result of this combination?

atombrella:
@nemesisdesign I found the comment somewhat confusing as well. In `TaggableModel` in `__get__`, an error is raised if the primary key is not set. However this is already being checked for in `test_require_pk` in `TaggableManagerTestCase`. Perhaps the two fragments can just be removed?

atombrella:
@frewsxcv Maybe you know the intend of those `TODO` test cases?

frewsxcv:
I'm not sure what those TODO comments are, to be honest. I haven't looked at the taggit source in a while so I'm not the best person to review this nowadays, but i think your changes are fine if you're comfortable with them.

atombrella:
I think those tests were what's intended. The `TODO` comments were unclear.

tests/tests.py
```diff
@@ -696,9 +696,10 @@ class TaggableManagerDirectCustomPKTestCase(TaggableManagerTestCase):
     taggeditem_model = TaggedCustomPKFood
 
     def test_require_pk(self):
-        # TODO with a charfield pk, pk is never None, so taggit has no way to
-        # tell if the instance is saved or not
-        pass
+        food_instance = self.food_model(name='foobar')
+        with self.assertRaises(self.food_model.DoesNotExist):
+            food_instance.refresh_from_db()
+            food_instance.pk
```
jdufresne: This line and the other `.pk` line are never executed as the previous line throws the exception. They can be removed.

jdufresne:
I think these new tests are mostly testing Django's `refresh_from_db()`. I'm unclear as to what aspect of taggit is being tested.

atombrella:
@jdufresne Maybe it's better to just update this PR to remove those tests marked with `TODO`?

jdufresne:
Yeah, that works for me. I think the intent of the TODO is lost to time.

jdufresne:
Thanks!

tests/tests.py
```diff
@@ -664,30 +667,6 @@ class TaggableManagerDirectTestCase(TaggableManagerTestCase):
     taggeditem_model = TaggedFood
 
 
-class TaggableManagerDirectCustomPKTestCase(TaggableManagerTestCase):
```
jdufresne: I think the test case should remain as it inherits several test functions that we still want to run.

I suppose this highlights the purpose of the TODO comment. It is overriding the test from the base class to avoid a failure. Is that right? If so, maybe it is sufficient to simply remove the word "TODO" but keep the remainder of the comment to explain why the test is skipped.

atombrella: Yes, there were failures when deleting those two tests.

PR #488, opened 2017-05-07
Use set literals and set comprehension

jdufresne:
Available since Python 2.7. Avoid some extra temporary lists and
function calls.


atombrella:
`flake8-comprehensions` <- perhaps you could include this in `tox.ini` in the same take?

PR #489, opened 2017-05-07
Add Django master to the test matrix

atombrella:
Adjust MIDDLEWARE/MIDDLEWARE_CLASSES with a conditional based on the version

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,7 @@
 language: python
 sudo: false
 matrix:
+  fast_finish: true
   include:
     - python: 2.7
       env: TOXENV=py27-1.8.x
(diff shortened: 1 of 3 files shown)


nemesisdesign:
It seems like a good idea. It could slow down the build a bit, but I've seen a similar approach in other widely used django packages. @frewsxcv what do you think?

frewsxcv:
SGTM, as long as `allow_failures` is `true`, which it appears to be

nemesisdesign:
`fast_finish` also allows to mark the build as success even if the tasks which are allowed to fail have not finished yet, I think it's pretty safe to merge. Thank you very much @atombrella 👍

PR #495, opened 2017-06-30
Fixed error when adding tags to user model with name of Name

tors10e:
The related_query_name function was using the model name instead of the related_name attribute which was causing a clash between the model name and the tag.name.  By returning the field's related name instead of the model name  when it exists, fixed this problem.


jdufresne:
Thanks! Could you add a test to avoid future regressions?

jdufresne:
I believe this was fixed in 0c47c93a73a174ab28542bae8db7d10ceb6a535b. Let me know if you continue to experience issues.

PR #500, opened 2017-09-28
small patch for managers.py

SkiBY:
Django 1.11.
Without name field in init we have an error: '_TaggableManager' object has no attribute 'name' 

class C(models.Model):
    ...
    tags = TaggableManager()
    ...

    def get_fields_with_value(self):
        return ((f, getattr(self, f, None)) for f in self.get_fields())


jdufresne:
Could you add a test to demonstrate the fix so we don't regress in the future?

rtpg:
this got fixed by having us call `__init__` in the `Manager` class, closing this out

PR #502, opened 2017-10-18
Enable pip cache in Travis CI

jdufresne:
Can speed up builds and reduce load on PyPI servers.

For more information, see:

https://docs.travis-ci.com/user/caching/#pip-cache

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: python
+cache: pip
 sudo: false
 matrix:
   fast_finish: true


nemesisdesign:
This looks good, I don't know why the build is failing though, doesn't seem related to this!

nemesisdesign:
Tried restarting the 3 failing jobs

jdufresne:
The Travis failures are occurring against the Django master branch only. Django master has some backwards incompatible changes that break the django-taggit tests. I think the master branch should be configured to allow failures as the master branch is not API stable and could introduce a backwards incompatible change at any time. I've fixed the Travis configuration in #504.

jdufresne:
Rebased. Tests now green.

PR #503, opened 2017-10-21
Add trove classifiers for general Python 2 & 3 support

jdufresne:


jdufresne:
The values are available at <https://pypi.python.org/pypi/django-taggit/json> it is possible that some scripts are using these values to perform reporting or other similar tasks regarding compatability. I sometimes do this.

Outside of scripting, I regularly use these values when assessing a package for use in a project. While I could infer Python 3 support from Python 3.6, I find the attention to detail an indicator of how a library is maintained and tested.

jdufresne:
Rebased. Tests now green.

PR #504, opened 2017-10-21
Allow Travis failures for Django master

jdufresne:
As Django master isn't fully stable, expect failures from time to time.


jdufresne:
Yup. Was dropped in https://github.com/django/django/commit/cfff2af02be40106d4759cc6f8bfa476ce82421c. Updated test matrix with this in mind.

nemesisdesign:
Thank you @timgraham @jdufresne @atombrella

PR #505, opened 2017-10-23
add renderer=None for django 2.1 compatibility

collinanderson:
Fixes TypeError: render() got an unexpected keyword argument 'renderer'

### DIFF
--- a/taggit/forms.py
+++ b/taggit/forms.py
@@ -8,7 +8,7 @@ from taggit.utils import edit_string_for_tags, parse_tags
 
 
 class TagWidget(forms.TextInput):
-    def render(self, name, value, attrs=None):
+    def render(self, name, value, attrs=None, renderer=None):
         if value is not None and not isinstance(value, six.string_types):
             value = edit_string_for_tags([
                 o.tag for o in value.select_related("tag")])


collinanderson:
I've rebased this change to the latest taggit-master. It fixes the issues with django-master, so it now passes.

collinanderson:
Hi All, Django 2.1 is now out. I've rebased my patch. (On Travis `isort says `models.py Imports are incorrectly sorted.`, but I didn't touch models.py)

gasman:
Ideally we should also pass on the `renderer` argument when calling `super`, so that it can be used for the intended purpose (namely, overriding the TextInput widget's template with a custom one using a different template engine):

`return super(TagWidget, self).render(name, value, attrs, renderer=renderer)`

(This will break on Django <1.11, though, so we'd need a version check if we're continuing to support those versions.)

Even better would be to move the `value`-rewriting logic to the [`format_value`](https://docs.djangoproject.com/en/2.1/ref/forms/widgets/#django.forms.Widget.format_value) method, so that we don't have to override `render` at all. (Again this needs a fallback for Django <1.10.) I'll open an alternative PR for this approach shortly.

collinanderson:
alternatively you could just accept `**kwargs` and then pass in `**kwargs` to super. that way you don't need a version check.

collinanderson:
Looks good to me. Thanks!

PR #506, opened 2017-11-01
Moved TaggedItemBase.tags_for() to ItemBase.

### LINKED ISSUE #101: type object 'TaggedSomething' has no attribute 'tags_for'
My class is inherited from `taggit.models.ItemBase`:

```
class TaggedSomething(ItemBase):
    content_object = models.ForeignKey('Something')
    tag = models.ForeignKey('MyTag')
```

However I got this error:
`type object 'TaggedSomething' has no attribute 'tags_for'.`

By digging the source it looks like `tags_for` is a mandatory class_method I need to implement. But I didn't see anywhere in the document. So I want to confirm it here.

pope1ni:
`ItemBase.tags_for()` was moved to `TaggedItemBase` in 998ef2f as part of a fix to avoid direct reference to the default `Tag` model. The change made use of `ItemBase.tag_model()` instead so moving `ItemBase.tags_for()` was unnecessary and makes subclassing `ItemBase` more difficult.

Fixes #101.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,11 @@
 Changelog
 =========
 
+UNRELEASED
+~~~~~~~~~~
+
+* Moved TaggedItemBase.tags_for() to ItemBase.
+
 1.1.0 (2019-03-22)
 ~~~~~~~~~~~~~~~~~~
 
--- a/taggit/models.py
+++ b/taggit/models.py
@@ -102,15 +102,6 @@ class ItemBase(models.Model):
     def lookup_kwargs(cls, instance):
         return {"content_object": instance}
 
-
-class TaggedItemBase(ItemBase):
-    tag = models.ForeignKey(
-        Tag, related_name="%(app_label)s_%(class)s_items", on_delete=models.CASCADE
-    )
-
-    class Meta:
-        abstract = True
-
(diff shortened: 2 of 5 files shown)


pope1ni:
Test failures are unrelated.

jdufresne:
Thanks! What do you think about adding test for this? It would avoid future regressions.

pope1ni:
Hi Jon. Have added some tests, as requested, for presence of `ItemBase.tags_for()` and two of the other `@classmethod`s on `ItemBase` that it uses, as well as a release note.

jdufresne:
I guess I was looking for a test that demonstrates how `ItemBase` should be used to build a custom tagged model. Something similar to the tests & models defined for `TaggedItemBase`

- https://github.com/alex/django-taggit/blob/05d2b1107b2c6ee56eed1405a7ba6a57bd6ae6a3/tests/models.py#L62-L93
- https://github.com/alex/django-taggit/blob/master/tests/tests.py#L660-L664

Beyond asserting that the method exists, it will show that it can be used without generating an error.

What do you think?

pope1ni:
Hi @jdufresne 

A year later and I have finally got back to this and added some proper tests as you requested.
Hopefully we can finally get this fix merged!

PR #512, opened 2017-12-13
Drop support for EOL Python 3.3

jdufresne:
Python 3.3 is EOL and no longer receiving bug fixes, including security fixes.

Python 3.3 is not supported by the latest pytest release, which is causing test failures on Travis and with tox.

https://docs.pytest.org/en/latest/changelog.html#pytest-3-3-0-2017-11-23


jdufresne:
Rebased. Tests now green.

PR #513, opened 2017-12-13
Remove runtests.py in favor of pytest

jdufresne:
The django-taggit suite should be run by using pytest. Having multiple test entry points that use different testing infrastructure is confusing to new contributors. As Travis and tox use pytest, standardize on that.


jdufresne:
Rebased. Tests now green.

PR #516, opened 2017-12-20
Add renderer argument to TagWidget to match Django widget interface

jdufresne:
Starting with Django 1.11, if a widget's `.render()` method does not have a renderer argument, a warning is issued of the form:

> RemovedInDjango21Warning: Add the `renderer` argument to the render() method of <class 'taggit.forms.TagWidget'>. It will be mandatory in Django 2.1.

Handle this argument to be future compatible and squelch the warning.

Added workaround for older Django versions as it is not always supported.

- Not supported: Django < 1.11
- Supported: Django 1.11 and 2.0
- Required: Django master

Tests against the master branch now pass.

For additional information see the Django release notes:

https://docs.djangoproject.com/en/1.11/releases/1.11/#id2

> The renderer argument is added to the `Widget.render()` method. Methods that don’t accept that argument will work through a deprecation period.

And the `Widget.render()` documentation:

https://docs.djangoproject.com/en/1.11/ref/forms/widgets/#django.forms.Widget.render

> Changed in Django 1.11:
>
> The renderer argument was added. Support for subclasses that don’t accept it will be removed in Django 2.1.

### DIFF
--- a/taggit/forms.py
+++ b/taggit/forms.py
@@ -2,17 +2,22 @@ from __future__ import unicode_literals
 
 from django import forms
 from django.utils import six
+from django.utils.inspect import func_supports_parameter
 from django.utils.translation import ugettext as _
 
 from taggit.utils import edit_string_for_tags, parse_tags
 
 
 class TagWidget(forms.TextInput):
-    def render(self, name, value, attrs=None):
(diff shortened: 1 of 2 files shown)


jdufresne:
If you agree with the approach taken in #518 then this PR can be closed.

PR #517, opened 2017-12-20
Sort imports per isort

jdufresne:


jdufresne:
Updated to sort imports per the new isort release.

PR #521, opened 2018-01-07
Added Finnish translation

sikmir:


sikmir:
@jdufresne Can we land all translations PRs #548, #522, #521?

jdufresne:
Thanks!

PR #525, opened 2018-01-27
Test against Django PyPI releases

jdufresne:
Helps ensure tests are running against released code that is being used
in production.

Allows local tox environments to used the local pip cache instead of
downloading a tarball each run. Reduces load on packaging servers and
network traffic.


tox.ini
```diff
@@ -9,11 +9,11 @@ envlist =
...
-    1.8.x: https://github.com/django/django/archive/stable/1.8.x.tar.gz#egg=django
-    1.9.x: https://github.com/django/django/archive/stable/1.9.x.tar.gz#egg=django
-    1.10.x: https://github.com/django/django/archive/stable/1.10.x.tar.gz#egg=django
-    1.11.x: https://github.com/django/django/archive/stable/1.11.x.tar.gz#egg=django
-    2.0.x: https://github.com/django/django/archive/stable/2.0.x.tar.gz#egg=django
+    1.8.x: Django>=1.8,<1.9
+    1.9.x: Django>=1.9,<1.10
+    1.10.x: Django>=1.10,<1.11
+    1.11.x: Django>=1.11,<2.0
+    2.0.x: Django>=2.0,<2.1
```
pope1ni: These can be simplified, e.g. `django~=2.0.0`. See [PEP440](https://www.python.org/dev/peps/pep-0440/) and [PEP508](https://www.python.org/dev/peps/pep-0508/).

jdufresne:
Thanks! I've made the suggested edit.

PR #534, opened 2018-08-02
Fix TagWidget for Django 2.1

gasman:
Proposed alternative fix to #505, which preserves the behaviour of the `renderer` kwarg.

`format_value` is a more suitable place for converting the value - this way we're sidestepping the need to deal with any current or future changes to Django's `render` logic.

(Unfortunately the Django <1.10 fallback code makes it pretty ugly, but hopefully that won't be there forever :-) )

### DIFF
--- a/taggit/forms.py
+++ b/taggit/forms.py
@@ -8,11 +8,12 @@ from taggit.utils import edit_string_for_tags, parse_tags
 
 
 class TagWidget(forms.TextInput):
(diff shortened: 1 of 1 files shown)


gasman:
@jdufresne Sure - rebased now. (Obviously, with #518 merged this change isn't really necessary any more, but I think performing the conversion in `format_value` is stylistically neater anyhow...)

jdufresne:
Thanks!

PR #551, opened 2018-11-13
Fix TaggableManager in Admin example code

stchris:
The current code throws an error:

```
(admin.E008) The value of 'fieldsets[0][1]['fields']' must be a list or tuple.
```


jdufresne:
Thanks!

PR #572, opened 2018-12-10
Adding slug convert option

hwshim0810:
Unidecode translates non-English words, I have added an option to use the original language as a slug.

I tried to add a test, but adding non-English characters to the test code seemed to be a problem.

### DIFF
--- a/docs/getting_started.txt
+++ b/docs/getting_started.txt
@@ -27,3 +27,15 @@ And then to any model you want tagging on do the following::
     If you want ``django-taggit`` to be **CASE INSENSITIVE** when looking up existing tags, you'll have to set to ``True`` the TAGGIT_CASE_INSENSITIVE setting (by default ``False``)::
 
       TAGGIT_CASE_INSENSITIVE = True
+
+.. note::
+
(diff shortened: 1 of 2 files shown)


jdufresne:
> I tried to add a test, but adding non-English characters to the test code seemed to be a problem.

Can you clarify? It should not be. We should add a test to verify the behavior before merging.

lucemia:
+1 for this PR. 

Without this PR, django-taggit has **serious** performance issue in any non-English applications.

https://github.com/jazzband/django-taggit/blob/83fe619d5b5f3b3eeb8524b36b066fb0dbbfe3f0/taggit/models.py#L61-L71

rtpg:
@lucemia thank you for pointing this out, I hadn't quite absorbed this fact when this had come up.

To be honest I don't quite know what the value of "ascii-fying" everything for slugs is, it does feel a bit silly. So I think that having this setting, _and_ having it be the new default (rationale: if you're using characters outside of ascii, it's very likely you find that important for your language).

rtpg:
Of course after thinking about this a bit more I am reminded that slugification exists primarily for URLs and the like.

For example if you have a URL scheme like `/posts/<tag>/`  then you want the tag to... at least not contain slashes I suppose.

I am going to try and dig up some ideas here for how to handle this.

PR #588, opened 2019-02-16
Make TaggedItem unique

### LINKED ISSUE #349: Race condition with get_or_create on TaggedItem

atugushev:
Prevents duplication of a`TaggedItem` on a race condition at this line:
https://github.com/jazzband/django-taggit/blob/7cea0b2d7af23d7c0e04f52d60abf3dd175cb26d/taggit/managers.py#L158-L160
 
The `kwargs` used in a `get_or_create` must be unique at the database level.

Fixes #349.

### DIFF
--- a/taggit/migrations/0003_taggeditem_add_unique_index.py
+++ b/taggit/migrations/0003_taggeditem_add_unique_index.py
@@ -6,11 +6,13 @@ from django.db import migrations
 
 class Migration(migrations.Migration):
 
-    dependencies = [("taggit", "0002_auto_20150616_2121")]
+    dependencies = [
+        ("contenttypes", "0002_remove_content_type_name"),
+        ("taggit", "0002_auto_20150616_2121"),
+    ]
 
     operations = [
         migrations.AlterUniqueTogether(
-            name="taggeditem",
-            unique_together=set([("content_type", "object_id", "tag_id")]),
+            name="taggeditem", unique_together={("content_type", "object_id", "tag")}
         )
     ]
--- a/taggit/migrations/0003_taggeditem_add_unique_index.py
+++ b/taggit/migrations/0003_taggeditem_add_unique_index.py
@@ -1,0 +1,16 @@
+# -*- coding: utf-8 -*-
+from __future__ import unicode_literals
+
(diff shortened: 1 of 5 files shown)


jdufresne:
Thanks for the PR! Can you port the test from #358 as well?

---

I think we'll need a data migration as well to handle existing duplicates in production databases. It will need to do something like:

- Find duplicates
- Look for FKs to `TaggedItem` and reassign them to the lowest PK
- Delete duplicates and keep the lowest PK

WDYT?

atugushev:
Hello @jdufresne,

> Thanks for the PR! Can you port the test from #358 as well?

I've ported a test and add @mythmon as co-author. Note that in the #358 the unique index in `GenericTaggedItemBase`, but in this PR the index added to specific model `TaggedItem`. That was done specially to not provocate migrations for inherited models on the user side. What do you think about that?

> I think we'll need a data migration as well to handle existing duplicates in production databases.

I agree with you, but on the other hand there might be a performance issue if a user have huge amount of tags. Do you think we should anyway provide a data migration?

jdufresne:
> Note that in the #358 the unique index in `GenericTaggedItemBase`, but in this PR the index added to specific model `TaggedItem`. That was done specially to not provocate migrations for inherited models on the user side. What do you think about that?

I agree. Good call.

> I agree with you, but on the other hand there might be a performance issue if a user have huge amount of tags. Do you think we should anyway provide a data migration?

Ok, we can leave this out for now. My concern is that if a database does have duplicates, this change will result in an `IntegrityError` when running migrate. I suppose we could add some strong words to the change log to help users out.

Issue #80, opened 2011-09-02
KeyError on similar_objects


karanbhangui:
```
Traceback:
File "/usr/local/lib/python2.7/dist-packages/django/core/handlers/base.py" in get_response
  111.                         response = callback(request, *callback_args, **callback_kwargs)
File "/usr/local/lib/python2.7/dist-packages/django/contrib/auth/decorators.py" in _wrapped_view
  23.                 return view_func(request, *args, **kwargs)
File "/home/ubuntu/web/notewagon-django/notewagon/views.py" in view_course
  120.     related = course.tags.similar_objects()
File "/usr/local/lib/python2.7/dist-packages/taggit/utils.py" in inner
  125.         return func(self, *args, **kwargs)
File "/usr/local/lib/python2.7/dist-packages/taggit/managers.py" in similar_objects
  230.                 tuple(result[k] for k in lookup_keys)

Exception Type: KeyError at /s/university-of-waterloo/fdsg1yl7dk/math-115-linear-algebra-eng
Exception Value: (35L, 1755L)
```

The steps to the trace above are as follows:
1. import/scrape a bunch of Resource items (I used youtube videos) and create tags based on the tags in the video
2. add some tags to a Course ("Linear Algebra", vectors, eigenvalues, etc)
3. observe error

The-Fonz:
I am experiencing the same error with `django-taggit 0.21.3`. The workaround doesn't work for me, it gives the same issue. Is there any other fix?

gelbander:
@The-Fonz Are you using SQLite? I'm having the same issue using sqlite (even with the workaround) but there is no issue with postgres.

The-Fonz:
@gelbander Yes I am, thanks for the tip, will try postgres!

The-Fonz:
I switched to postgres but still had the KeyError, fixed it by changing line 260 in `managers.py` from `items[(getattr(obj, remote_field.field_name),)] = obj` to `items[(obj.id,)] = obj`, so now line 275 can properly access the dict. I'm using Wagtail but I'm not deep enough into Wagtail/Django/ModelCluster/Taggit to see why this bug appears. I implemented the fix by subclassing `_ClusterTaggableManager` and `ClusterTaggableManager`.

return1:
i am having the same issue with @django-taggit 0.22.1@ using Wagtail CMS

strindhaug:
[...]

PR #599, opened 2019-03-14
Workaround KeyError in similar_objects

jaap3:
This fixes the symptom, but not the cause. 

Since this entire method is deemed to be "a bit of a hack", and the issue has been open since 2011,--and is possibly cause by a Django issue--this seems like a sufficient solution.

I've successfully applied this patch in production (by extracting this method into a get_similar_objects function). The method will still return similar_objects, but no longer crashes on missing keys.

Ref: #80

### DIFF
--- a/taggit/managers.py
+++ b/taggit/managers.py
@@ -377,9 +377,7 @@ class _TaggableManager(models.Manager):
         results = []
         for result in qs:
             try:
-                obj = items[
-                    tuple(result[k] for k in lookup_keys)
-                ]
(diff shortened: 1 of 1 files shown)


jdufresne:
Thanks! Can you please add a test for the issue?

jaap3:
I don't know the root cause of this issue. This makes it hard (impossible) to write a tests that creates the scenario this patch is supposed to work around.

jdufresne:
Sorry, but I'd prefer not to merge without a test that demonstrates the fix and prevents regressions.

PR #625, opened 2019-05-30
Change deprecated private attribute to support Django>=1.11

DmytroLitvinov:
Hello @jdufresne!

Closes https://github.com/jazzband/django-taggit/issues/624

### DIFF
--- a/taggit/managers.py
+++ b/taggit/managers.py
@@ -547,7 +547,7 @@ class TaggableManager(RelatedField):
 
     def m2m_column_name(self):
         if self.use_gfk:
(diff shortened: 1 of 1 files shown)


auvipy:
thanks

jdufresne:
Thanks @DmytroLitvinov! Is it possible to add a test that exercises this code?

DmytroLitvinov:
@jdufresne I am not sure.
Need to check.

jdufresne:
Thanks, that would be very helpful to avoid future regressions. A good place to start would be to mimic the code path that triggered this bug for you in the first place.

PR #630, opened 2019-08-07
Add support for `RelatedManager.add()` `through_defaults` (Django 2.2)

soulshake:
This is a proof of concept intended to address https://github.com/jazzband/django-taggit/issues/629.

Problems:

- I don't think this change would be compatible with Django < 2.2, at least not in its current state.
- I am not sure how to test this properly (I made an attempt but haven't gotten it working).

I would be happy to put some more work into this if it's a desired change. If so, I would appreciate some guidance on how to go about testing it.


docs/api.rst
```diff
@@ -21,6 +21,9 @@ playing around with the API.
             []
             >>> apple.tags.add("red", "green", "fruit")
 
+        Use the `through_defaults` argument to specify values for your custom
+        `through` model, if needed.
```
jdufresne:
```suggestion
        ``through`` model, if needed.
```

docs/api.rst
```diff
@@ -12,7 +12,7 @@ playing around with the API.
         information.
     :param blank: Controls whether this field is required.
 
-    .. method:: add(*tags)
+    .. method:: add(*tags, through_defaults={})
```
jdufresne:
```suggestion
    .. method:: add(*tags, through_defaults=None)
```

jdufresne:
Hi @soulshake Thanks for the PR. Are you interested in cleaning up the failing tests?

soulshake:
@jdufresne I would be, but I wasn't sure how to go about it.

PeteAndersen:
I did some digging - each of the test case classes use specific models to test. The tests for using custom through models all use the `OfficialFood` model, and the unit test for testing `through_defaults` should be in the `TaggableManagerOfficialTestCase` test class.

PeteAndersen:
Testing with Django 1.11 seems to work fine, since the `through_defaults` are passed into `get_or_create()` as the `defaults` kwarg, which has support across all the targeted Django versions.

PeteAndersen:
It would be good to change the function signature for `add()` so that through_defaults is pulled out of keyword args. With it defined as a positional argument you would have to explicitly pass in None when using any kwargs, such as in #653

auvipy:
please fix the merge conflict

taggit/managers.py
```diff
@@ -143,7 +143,7 @@ def _lookup_kwargs(self):
         return self.through.lookup_kwargs(self.instance)
 
     @require_instance_manager
-    def add(self, *tags, **kwargs):
+    def add(self, through_defaults=None *tags, **kwargs):
```
auvipy, line 146: missing comma!

PR #632, opened 2019-08-19
modify contribute_to_class method,so it can use in abstract class

xunull:
I add tags field in a abstract model class, but where I  run makemigrations,it has some error about reverse related_name, so I used some code which in RelatedField's contribute_to_class method to resolve the error


xunull:
> need tests

i have changed some code according to black

auvipy (dismissed):

jdufresne:
This still does not have a test. Could you add one please? A test will help demonstrate its usefulness and avoid regression. Thanks.

PR #653, opened 2019-11-26
Allow for passing in extra kwargs on Tag creation

rtpg:
When [using custom tags](https://django-taggit.readthedocs.io/en/latest/custom_tagging.html#using-a-custom-tag-or-through-model), you might need to pass in extra keyword arguments when creating new tags through the model manager.

This code change adds the ability to do so, through a `tag_kwargs` argument in the `add` method of the tag managers.

Example usage (if you were using a custom tag model with a `tenant` key, for multi-tenant applications):

```
        document.tags.add("a new tag", tag_kwargs={'tenant': self.tenant})
```

Sending this in as-is for comment, though we have been using these changes in production for quite a long time at this point so at least for our use case it's been valuable.

 If this looks like the kind of thing that y'all want, I can add documentation and tests for this.


PeteAndersen:
This type of change is definitely needed to make custom tag model usage practical.

jdufresne:
How does this compare to #630? Are these duplicating one another?

PeteAndersen:
`through_defaults` targets extra fields on the intermediate m2m model, while `tag_defaults` would target extra fields on the tag model. I think they're both useful to support.

rtpg:
Alright, thanks for the feedback.  Normally we shouldn't need any more migrations, but we should add some documentation + tests for this. I'll see what I can do here

rtpg:
Sorry for the wait, I've added (meager) documentation and a test for this. I think this is good to go in principle though. @auvipy are there any other changes that you would like to see for this?

StriveForBest:
is it supposed to work? I am getting an error:
```
TypeError: add() got an unexpected keyword argument 'tag_kwargs'
```

rtpg:
@StriveForBest I do not believe there has been a release since this merge, so unless you're installing a specific targetted commit you might not have this change

rtpg:
@StriveForBest we just released a new version of this (`1.3.0`), you should be able to use this feature now

PR #666, opened 2020-03-22
docs: Fix simple typo, oportunistic -> opportunistic

timgates42:
There is a small typo in taggit/models.py.

Should read `opportunistic` rather than `oportunistic`.


jdufresne:
Thank you!

PR #673, opened 2020-05-19
Prepare for the next release

rtpg:
I'm not sure who is the release manager for this lib, and I don't see any "autorelease on tag" infrastructure, so I'm going to ping both @auvipy and @jdufresne on this (apologies in advance to the one who didn't need to get pinged here)

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,13 +1,17 @@
 Changelog
 =========
 
-UNRELEASED
-~~~~~~~~~~
+1.3.0 (2020-05-19)
+~~~~~~~~~~~~~~~~~~
 
 * Model and field ``verbose_name`` and ``verbose_name_plural`` attributes are
   now lowercase. This simplifies using the name in the middle of a sentence.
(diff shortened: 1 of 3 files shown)


CHANGELOG.rst
```diff
@@ -1,13 +1,17 @@
 Changelog
 =========
 
-UNRELEASED
-~~~~~~~~~~
+1.2.1 (2020-05-19)
```
jdufresne, line 4: I think we should go to 1.3.0. WDYT?

rtpg: Hesitated on this point, let's go with  1.3.0! I'll send in the changes

rtpg: Alright 1.3.0 it has become

jdufresne:
Thanks. Please feel free to merge and release at your convenience.

rtpg:
Alright I pushed up the tag, my understanding is that this should trigger something over on your end as release managers. Thanks for the super fast response times :100:

PR #681, opened 2020-07-30
Add tip for using custom models

steezeburger:
If you are using custom models, having 'taggit' in the installed apps will create database tables you do not need.

### DIFF
--- a/docs/custom_tagging.rst
+++ b/docs/custom_tagging.rst
@@ -12,7 +12,9 @@ want to store additional data about a tag, such as whether it is official.  In
 these cases ``django-taggit`` makes it easy to substitute your own through
 model, or ``Tag`` model.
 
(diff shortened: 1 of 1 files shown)


docs/custom_tagging.rst
```diff
@@ -12,6 +12,8 @@ want to store additional data about a tag, such as whether it is official.  In
 these cases ``django-taggit`` makes it easy to substitute your own through
 model, or ``Tag`` model.
 
+Note: You will need to remove 'taggit' from `settings.py` if you do not want to create the default tables.
```
auvipy, line 15: can you be more specific? the sentence needs more detail and context.

steezeburger: @auvipy I've updated the Note. Let me know if it needs more detail!

PR #693, opened 2020-09-30
Allowing BigIntegerField in taggedItems

sowdust:
Allows Taggable Items to have BigIntegerField as primary key. See https://github.com/jazzband/django-taggit/issues/692

### DIFF
--- a/taggit/models.py
+++ b/taggit/models.py
@@ -157,7 +157,7 @@ class CommonGenericTaggedItemBase(ItemBase):
 
 
 class GenericTaggedItemBase(CommonGenericTaggedItemBase):
-    object_id = models.IntegerField(verbose_name=_("object ID"), db_index=True)
+    object_id = models.BigIntegerField(verbose_name=_("object ID"), db_index=True)
 
     class Meta:
         abstract = True


taggit/models.py
```diff
@@ -157,7 +157,7 @@ def tags_for(cls, model, instance=None, **extra_filters):
 
 
 class GenericTaggedItemBase(CommonGenericTaggedItemBase):
-    object_id = models.IntegerField(verbose_name=_("object ID"), db_index=True)
+    object_id = models.BigIntegerField(verbose_name=_("object ID"), db_index=True)
```
auvipy, line 160: this will break backward compatibility

auvipy (changes requested):

taggit/models.py
sowdust: I see no other way to fix this issue  : (

gladson: They could do a version check of Django before migrating.
And use:

```shell
In [1]: import django
In [2]: django.VERSION
Out [2]: (3, 1, 1, 'final', 0)
```

auvipy: probably moving to abstract base model is the best futuristic approach

rtpg:
This change as-is would break backwards compatibility with everyone using this base. After triaging a hundred+ issues, I think a big thing is the custom tag documentation. I'm going to work on improving it so that there could be a good answer here. But this as-is won't work, unfortunately

PR #698, opened 2020-10-25
Danish translation added

jakobdo:


rtpg:
(Added a changelog entry for this + something esle that was missing a changelog entry, will merge this in once it passes)

rtpg:
Thank you for these translations! They'll be in the next release

Issue #508, opened 2017-11-12
KeyError with `similar_objects()` function


DDiran:
Hi Alex,

I'm working on a project using `Wagtail-CMS` which incorporates your tagging system. For what I'm working on at the moment the `similar_objects()` function you created is exactly what I need, however I think there might be an issue with the function (unless I'm making some silly mistake in my own code).

I've written up a detailed post on SO which you can check out [here](https://stackoverflow.com/questions/47240510/getting-related-objects-via-tags-in-django).

Many thanks

prokaktus:
Stuck with the same problem. Eventually I decided to use `raw` sql to perform my query. I wrote this way (It is just **workaround**, not solution to original problem.):
```
ArticlePage.objects.raw('select a.page_ptr_id, p.title, count(at.tag_id) as tag_count from article_articlepage a join wagtailcore_page p on a.page_ptr_id = p.id join article_articletag at on at.content_object_id = a.page_ptr_id join taggit_tag t on t.id = at.tag_id where tag_id in (1, 2, 3) group by (page_ptr_id, p.id) order by tag_count desc');
```

I put this into gist with better explanation, so feel free to copy and modify it if you needed: https://gist.github.com/prokaktus/bd805e3f9a3eabb64be22523ae5fdf66

legshort:
I'm having the same issue and seems like this pr #424 is kind of stuck for the last few years.

jaap3:
This seems to be a duplicate of #80

jdufresne:
Thanks.

PR #712, opened 2021-03-02
Fix similar_objects() in multi inheritance contexts (KeyError)

czr137:
This addresses issues wagtail users have had using `similar_objects()` with `ClusterTaggableManager`

This PR would resolve both #424 and #508 and close issue #80.

It is an updated and rebased version of the commits made in #424 by nickhudkins.

It finishes the work in PR #424 in that drops the workarounds for older versions of Django as requested by @jdufresne in #424.


czr137:
At first glance it looks like a problem with the test scripts for 3.6 and 3.7 and not an issue with this PR. Can a JazzBand maintainer confirm?

czr137:
Test scripts for 3.6 and 3.7 are attempting to use a version of Django that no longer supports python 3.6 and 3.7. Looks like these test builds could be dropped, or they need to be edited to specify an older version of Django.

tests/models.py
```diff
@@ -57,6 +57,29 @@ def __str__(self):
...
+        return self.name
+
+
+class MultiInheritanceLazyResolutionFoodTag(TaggedItemBase):
+    content_object = models.ForeignKey(
+        "MultiInheritanceFood", related_name="tagged_items", on_delete=models.CASCADE
+    )
+
+    class Meta:
+        unique_together = [["content_object", "tag"]]
```
auvipy, line 73: instead of unique_together you could try UniqueContraint

auvipy: https://docs.djangoproject.com/en/2.2/ref/models/options/#unique-together

czr137: There are about a half dozen occurrences of unique_together in this file, and none of them have been switched over to UniqueConstraint yet. I think for clarity, as this would be an unrelated change, it would be better to change all the unique-together to UniqueConstraint with its own issue / PR.

czr137: UniqueConstraint was introduced in Django 2.2, so if I'm not mistaken, this would also mean the min version of Django would have to be bumped to 2.2 and version 1 support would have to be dropped. This is probably fine at this point, but it is also beyond the scope of this PR.

czr137:
Django 4.0 has dropped support for python 3.6,3.7.
As such, I modified the tox test script so that it doesn't test python 3.6 and 3.7 against django master, which has been updated to support >=3.8.

tests/migrations/0001_initial.py
```diff
@@ -31,6 +31,21 @@ class Migration(migrations.Migration):
                 ("title", models.CharField(max_length=100)),
             ],
         ),
+        migrations.CreateModel(
```
auvipy, line 34: did you created the migrations yourselves? are they going to break the downward migrations?

czr137: > did you created the migrations yourselves?

Yes.

> are they going to break the downward migrations?

Please review them. The tests are passing.

czr137:
@auvipy, please let me know if there is anything else I can do to help get this fix in. If there is some other way to demonstrate the migrations I added are correct, let me know and I'll try to do that over the weekend.

czr137:
@auvipy, I removed my manual migrations and added a migrations file with django's generated migration. I rebased onto @jezdez test updates so the two PR's are consitent.
Tests are all passing and test migrations are fixed.

auvipy (approved):

czr137:
#599 can also be closed when this is merged in.

jdufresne:
@auvipy I'm no longer actively using or maintaining django-taggit. Please feel free to review, merge, and release as you see fit.

auvipy (approved):

PR #720, opened 2021-04-14
Update django support

hramezani:


CHANGELOG.rst
```diff
@@ -8,6 +8,8 @@ Unreleased
 * Dropped support for Python 3.5 and Django 1.11.
 * Fix crashing that could occur with ``similar_objects`` in multi-inheritance contexts.
 * Add Danish translation.
+* Remove Django 3.0 support.
```
auvipy, line 11: support of django 3 can't be dropped unill django 2.2 retain. it is only removed from marix

rtpg:
read through the 3.2 release notes and nothing stood out to me, so this looks good. Thanks for sending this in!

auvipy (approved):

CHANGELOG.rst
rtpg: I thought Django 3.0 was out of support (whereas 2.2 is still LTS). I guess Django 3.0 will continue to work for a long time given how this stuff works and its proximity to 2.2, though. Am I missing some detail?

auvipy: > I guess Django 3.0 will continue to work for a long time given how this stuff works and its proximity to 2.2, though

That is what I wanted to mean. no problem.

PR #733, opened 2021-06-10
Update versions

hramezani:


rtpg (approved):
Alright, looks like we weren't even testing 1.11 at this point anyways. Thanks for updating the docs

PR #735, opened 2021-06-16
Support ordering tags by specified field

kavdev:
A starting point to close https://github.com/jazzband/django-taggit/issues/370.

It's currently a bit tricky to get ordered tags (unless I'm missing something). This enables specifying an ``ordering`` argument to the manager's init that causes tags to be pulled from the database in a user-specified order.


auvipy (approved):

auvipy (changes requested):
can you add test and check the failures?

auvipy (approved):

PR #738, opened 2021-06-19
Vendor in `django-taggit-serializer`

rtpg:
This is after attempting to contact the original repo owner and not getting any replies.

The licenses are identitcal, so vendoring shouldn't be a problem on that front. 

I also moved this into `taggit.serializers` so that people can transition smoothly, or at their own pace/never.


rtpg:
Alright this lowers overall project coverage, but this was mainly about vendoring in the existing project so I'm going to merge this in as-is. 

I plan on doing some cleanup for the serializers at a later point (as I think the current mechanism is more frail than it needs to be given DRF's capabilities), but I'll pull this in and just prepare a release.

taggit/serializers.py
```diff
@@ -0,0 +1,123 @@
+"""
+Django-taggit serializer support
+
+Originally vendored from https://github.com/glemmaPaul/django-taggit-serializer
+"""
+import json
+
+# Third party
+from django.utils.translation import ugettext_lazy as _
```
auvipy, line 9: ugettext_lazy could bee gettext_lazy

taggit/serializers.py
```diff
@@ -0,0 +1,123 @@
...
+from rest_framework import serializers
+
+
+class TagList(list):
+    def __init__(self, *args, **kwargs):
+        pretty_print = kwargs.pop("pretty_print", True)
+        list.__init__(self, *args, **kwargs)
```
auvipy, line 16: what about using super call?

auvipy (commented):
good job

PR #740, opened 2021-06-30
Release Cleanup + Preparing for Release 1.5.0

rtpg:


rtpg:
The failing coverage thing is about a non-covered line still not being covered. I think we should increase coverage of this file overall, but this is meant to be a vendoring of the project for now so I'm going to move forward with this as-is.

PR #756, opened 2021-09-30
Fix Ukrainian language definition

sergeyklay:


sergeyklay:
@null-none Could you please take a look

Issue #686, opened 2020-08-20
TaggableManager.set() API does not match RelatedManager.set()


rjmackay:
TaggableManager has `set(*tags, clear=False)` while RelatedManager expects the tag objects as a list ie `set(tags, bulk=True, clear=False, through_defaults=None)`.
Would it be possible to support setting tags as a list/tuple too? The different interface makes it harder to plug that TaggableManager into other django libraries.

auvipy:
feel free to dig in and start a WIP PR to improve it?

rtpg:
I'm tempted to qualify this as a bug, due to not following SOLID. Thanks for pointing this out. A simple base case here would be to take params, but throw a `NotImplementedError` on stuff we don't support.

coredumperror:
I agree 100% that this is a bug. I've now been burned by this difference in APIs between `TaggableManager.set()` and `RelatedManager.set()`, as I'm trying to use the same code to set various different types of tags on objects, and only one of those types is handled by django-taggit. So the API being different screws me.

rtpg:
@coredumperror feel free to send in a PR, I'll gladly review it to get it in. I think when I made the original comment I didn't see a straightforward way of getting it to work

pablolmedorado:
I've been facing this issue trying to use a tags field together with the library "[django-import-export](https://github.com/django-import-export/django-import-export)". It's been impossible.

PR #763, opened 2021-11-09
Make TaggableManager.set API compatible with Django's RelatedManager.set

pablolmedorado:
See #686


auvipy (changes requested):
please fix the conflicts and rebase issues

docs/api.rst
```diff
@@ -36,7 +36,7 @@ playing around with the API.
 
         Removes all tags from an object.
 
-    .. method:: set(*tags, through_defaults=None, clear=False)
+    .. method:: set(tags, through_defaults=None, clear=False)
```
auvipy, line 39: def set(self, tags, *, through_defaults=None, **kwargs) has separate *

pablolmedorado: Done! Nice spot

rtpg:
I don't consider this a blocker but I think it would be good if the release notes just went into more detail, saying that this changed from taking varargs to taking an explicit list (perhaps just an example). Without that people will have to dig into the code to figure out what change is needed here.

pablolmedorado:
I agree and am willing to change it. The reason why I have not done it from the beginning is because I am not very confident in my level of English. It is my first pull request and I did not know how to proceed.

If you want, I could try to explain it and, if you think there is a better translation, you can correct me.

Best regards.

auvipy:
> I agree and am willing to change it. The reason why I have not done it from the beginning is because I am not very confident in my level of English. It is my first pull request and I did not know how to proceed.
> 
> If you want, I could try to explain it and, if you think there is a better translation, you can correct me.
> 
> Best regards.

sure please

rtpg:
@pablolmedorado that sounds good, yes! It doesn't have to be very long, just a bit more details (given this would be a backwards incompatible change)

Given that it's been a while since the last release, I think once this is cleaned up we can package up a release

pablolmedorado:
I've updated the changelog according to what you pointed out yesterday. Feel free to modify anything you consider wrong.

Thanks for your patience! BR.

auvipy (approved):

auvipy (approved):

PR #766, opened 2021-11-14
Fix up translations based on recent code changes

rtpg:
Two code changes happened recently that weren't handled in most
translations:

- the lowercase-ification of content types
- the adding of a context to the "slug" and "name" taggit admin thingy

This commit just goes in and assumes that this doesn't affect other
languages' translations. In reality it's totally possible that some
stuff should be lowercased or otherwised changed, but I'm making a
judgement call that the same translations for the text as before will
work well.

This also introduces empty translations for missing ones, which should
help future translators see what is missing.


rtpg:
(This was a release blocker I noticed reviewing the code changes between 1.5.1 and now)

auvipy (approved):

PR #767, opened 2021-11-14
Prepare release

rtpg:
This is version 2.0.0 because of an API compatibility change.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,9 @@ Changelog
 (Unreleased)
 ~~~~~~~~~~~~
 
+2.0.0 (2021-11-14)
+~~~~~~~~~~~~~~~~~~
+
 * **Backwards incompatible:** ``TaggableManager.set`` now takes a list of tags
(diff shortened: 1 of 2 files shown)


jezdez:
@rtpg This should have been merged before the 2.0.0 tag was tagged so the changelog shows up correct only Read The Docs.

rtpg:
Alright! So from now on for releases I'll do a prep commit, merge that in, then tag the merge commit. 

@jezdez should I delete and retag the 2.0.0 release then? I haven't pushed the release out yet so I feel like we could still do that safely

jezdez:
> Alright! So from now on for releases I'll do a prep commit, merge that in, then tag the merge commit.
> 
> @jezdez should I delete and retag the 2.0.0 release then? I haven't pushed the release out yet so I feel like we could still do that safely

Yeah, that's not a bad idea, make sure to delete the files on [jazzband.co](https://jazzband.co/projects/django-taggit#2.0.0) first though. And let me know if you have trouble with the tag deletion :)

PR #777, opened 2021-12-15
Add Django 4.0 to testing environments

monty5811:
Just adds Django 4.0 to the ci

Currently fails due to (at least) #776.

Issue #776, opened 2021-12-07
`get_extra_restriction()` missing required positional argument related_alias


bsiebens:
Using Django-Taggit on Django 4.0rc1 (running on Python 3.9). I have a model called TransactionJournal that contains following line:

`tags = TaggableManager(blank=True)`

Querying the model with the following statement results in the below error:

`TransactionJournal.objects.filter(tags__name__in=["t1", "t2"])`

`Traceback (most recent call last):
  File "<console>", line 1, in <module>
  File "/Users/XXX/.local/share/virtualenvs/YYY-rEzd9bfd/lib/python3.9/site-packages/django/db/models/query.py", line 256, in __repr__
    data = list(self[:REPR_OUTPUT_SIZE + 1])
  File "/Users/XXX/.local/share/virtualenvs/YYY-rEzd9bfd/lib/python3.9/site-packages/django/db/models/query.py", line 280, in __iter__
    self._fetch_all()
  File "/Users/XXX/.local/share/virtualenvs/YYY-rEzd9bfd/lib/python3.9/site-packages/django/db/models/query.py", line 1354, in _fetch_all
    self._result_cache = list(self._iterable_class(self))
  File "/Users/XXX/.local/share/virtualenvs/YYY-rEzd9bfd/lib/python3.9/site-packages/django/db/models/query.py", line 51, in __iter__
    results = compiler.execute_sql(chunked_fetch=self.chunked_fetch, chunk_size=self.chunk_size)
  File "/Users/XXX/.local/share/virtualenvs/YYY-rEzd9bfd/lib/python3.9/site-packages/django/db/models/sql/compiler.py", line 1189, in execute_sql
    sql, params = self.as_sql()
  File "/Users/XXX/.local/share/virtualenvs/YYY-rEzd9bfd/lib/python3.9/site-packages/django/db/models/sql/compiler.py", line 545, in as_sql
    from_, f_params = self.get_from_clause()
  File "/Users/XXX/.local/share/virtualenvs/YYY-rEzd9bfd/lib/python3.9/site-packages/django/db/models/sql/compiler.py", line 833, in get_from_clause
    clause_sql, clause_params = self.compile(from_clause)
  File "/Users/XXX/.local/share/virtualenvs/YYY-rEzd9bfd/lib/python3.9/site-packages/django/db/models/sql/compiler.py", line 463, in compile
    sql, params = node.as_sql(self, self.connection)
[...]

PR #778, opened 2021-12-15
Fix Django 4 compatibility issues

monty5811:
The `where_class` argument to the private `get_extra_restriction()` was [changed](https://docs.djangoproject.com/en/4.0/releases/4.0/#miscellaneous) in Django 4.0.

A simple change is to remove the `where_class` argument [here](https://github.com/jazzband/django-taggit/blob/4042b8562d72b8b65581de01277e8b37514cf514/taggit/managers.py#L669). But that breaks taggit on lower versions of django.

To fix this, I added a shim that checks the django version and uses the correct function. 
I have no idea if this is the best way to fix this, but the tests pass 🤷‍♂️ 

Relevant issue: #776 

Note that this PR has a commit from #777 too.

### DIFF
--- a/taggit/managers.py
+++ b/taggit/managers.py
@@ -1,6 +1,7 @@
 import uuid
 from operator import attrgetter
 
+import django
 from django.conf import settings
 from django.contrib.contenttypes.fields import GenericRelation
 from django.contrib.contenttypes.models import ContentType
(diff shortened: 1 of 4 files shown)


taggit/managers.py
```diff
@@ -666,14 +667,26 @@ def get_joining_columns(self, reverse_join=False):
...
         return ExtraJoinRestriction(related_alias, extra_col, content_type_ids)
 
+    def _get_extra_restriction_legacy(self, where_class, alias, related_alias):
+        # this is a shim to maintain compatibility with django < 4.0
+        return self._get_extra_restriction(alias, related_alias)
+
+    # this is required to handle a change in Django 4.0
+    # https://docs.djangoproject.com/en/4.0/releases/4.0/#miscellaneous
+    # the signature of the (private) funtion was changed
+    if django.VERSION > (3, 8):
```
auvipy, line 685: 3, 8 or 3, 2?

monty5811: Ah, woops, should be `(3, 2)`.

hramezani:
Thanks @monty5811 :+1: 

- Please add `Django 4.0` to [setup classifiers](https://github.com/jazzband/django-taggit/blob/4042b8562d72b8b65581de01277e8b37514cf514/setup.cfg#L10)
- Add a changelog entry to `CHANGELOG.rst`

rtpg:
Looks good to me too, once we fix up the above comments! Thank you for sending this in :+1:

auvipy (approved):

hramezani (approved):
LGTM

taggit/managers.py
monty5811: Fixed this, I flipped it round to do ` < (4, 0)` instead.

steven-mercatante:
Hi, when can we expect a new version including this fix to land in pypi?
`v2.0.0`'s last commit was November 16, 2021. Thanks!

PR #779, opened 2021-12-16
Update django

hramezani:
Django 4.0 is out and it's time to drop support for all versions older than the most recent LTS.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,7 @@ Changelog
 
 * Add Python 3.10 support.
 * Add Django 4.0 support.
+* Drop Django 3.1 support.
 
 2.0.0 (2021-11-14)
 ~~~~~~~~~~~~~~~~~~
--- a/setup.cfg
+++ b/setup.cfg
@@ -12,7 +12,6 @@ classifiers =
     Environment :: Web Environment
     Framework :: Django
     Framework :: Django :: 2.2
-    Framework :: Django :: 3.1
     Framework :: Django :: 3.2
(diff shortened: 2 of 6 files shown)


hramezani:
I think we can prepare a release after we merged this.

README.rst
```diff
@@ -57,7 +57,7 @@ Then you can use the API like so:
 
 Tags will show up for you automatically in forms and the admin.
 
-``django-taggit`` requires Django 2.2 or greater.
```
auvipy, line 60: i think django 2.2 is still supported til april 2022

auvipy (changes requested):
except python 3.6+ , django should remain 2.2+ for next april

README.rst
hramezani: I was about to keep Django 2.2 til April 2022. but here is Tim [comment](https://github.com/jazzband/django-hosts/pull/141#discussion_r766116557) about it:

> No, once Django 4.0 is out, it's time to drop support for all versions older than the most recent LTS. Otherwise, it could require ugly shims to add 4.0 support (which removes a lot of deprecated things). This is outlined in https://www.djangoproject.com/weblog/2015/jun/25/roadmap/.

auvipy: but for smoother upgrade path keeping compat little bit longer is good for users

hramezani: Thanks @auvipy for the review. I updated the PR and removed just Django 3.1

auvipy: great thanks

auvipy (approved):

PR #792, opened 2022-02-22
Handle path_infos / reverse_path_infos cached properties (Dj4.1 compatibility)

gasman:
In https://github.com/django/django/commit/a697424969f7f464bf6492b09a6cdac135499e02 / https://github.com/django/django/pull/14750, Django has introduced two new cached properties `path_infos` and `reverse_path_infos` to be used in preference to the `get_path_info()` and `get_reverse_path_info()` methods when not passing a `filtered_relation` argument.

TaggableManager is patching these methods, so to avoid Django bypassing them (specifically in `django.db.query.Query.names_to_path`) it needs to provide these cached properties too.

For additional bonus points, we also update TaggableManager's `_get_mm_case_path_info` and `_get_gfk_case_path_info` methods to make use of the new cached properties where available.

This (along with #791, incorporated here) gets the test suite passing again on the current Django main branch.

### DIFF
--- a/taggit/managers.py
+++ b/taggit/managers.py
@@ -14,6 +14,7 @@ from django.db.models.fields.related import (
     lazy_related_operation,
 )
 from django.db.models.query_utils import PathInfo
+from django.utils.functional import cached_property
 from django.utils.text import capfirst
 from django.utils.translation import gettext_lazy as _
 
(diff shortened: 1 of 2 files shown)


gasman:
@rtpg Thanks! I don't think there's any particular hurry to get a release out with this fix - at least until Django 4.1 is in release candidate stage, which is still a couple of months away (final release is scheduled for August). My motivation for fixing this now is to [keep Wagtail's test suite passing against Django main](https://github.com/wagtail/wagtail/pull/8028) so that we get early warning of any upcoming breaking changes - and for that purpose it's no problem for us to test against django-taggit master.

PR #793, opened 2022-03-18
Docs: replace deprecated/removed `django.utils.translation.ugettext_lazy`

Abbe98:
`django.utils.translation.ugettext_lazy` was deprecated in Django 3.0 and removed in Django 4.0 in favor of `django.utils.translation.gettext_lazy`.


hramezani:
Thanks @Abbe98

PR #797, opened 2022-04-21
Preserve unicode when slugifying by default

rtpg:
Thanks to @hwshim810 (for original implementation) and @lucemia 
(for ping + pointing out the perf weirdness) from #572 for motivating this.
 
 This change is motivated by the many many people using non-latin
languages who experience the (rather baffling) behavior of outright
stripping characters when generating slugs from stuff that doesn't
fit into ASCII.

 We went through loads of pain as a programming community to get to
nicely supporting unicode everywhere, it's time to take advantage of
that fact and just let people have stuff appear in their native
language as much as possible.

PR #798, opened 2022-04-24
set allow_unicode True for slugs

gitemad:
set allow_unicode True for slugs to use unicode characters in slugs.


rtpg:
resolved this in #797

PR #800, opened 2022-04-28
Version bump to 3.0.0

gasman:
Hopefully not controversial (#797 is an upcoming breaking change, and django-taggit has followed semver in previous cases like this) but if you're curious why I'm proposing this right now: we're [testing Wagtail against Django git main](https://github.com/wagtail/wagtail/blob/8efb235695887ca6b0388f7f5640452ee92a2c22/.github/workflows/test.yml#L82-L85), which necessitates also installing django-taggit git master for the 4.1 fixes (#791, #792). Our test suite includes a check for missing migrations, and currently that's being triggered by the `allow_unicode` change in #797. Bumping the version now will allow us to hack a version check into our migrations and keep them working against both the current release version and git master.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,8 +1,8 @@
 Changelog
 =========
 
-(Unreleased)
-~~~~~~~~~~~~
+3.0.0 (Unreleased)
+~~~~~~~~~~~~~~~~~~
 
 * **Backwards incompatible:** Tag slugification used to silently strip non-ASCII characters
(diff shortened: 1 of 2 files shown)


rtpg:
Yep, I was thinking of setting up a new release. The unicode slugification change in particular is a big breaking change.

I am going to figure out how to get get us to a release here...

rtpg:
@gasman I've gone and made a full release here. Closing this PR as I believe it is no longer needed

PR #802, opened 2022-05-03
Clear prefetch cache on tag updates

hartungstenio:
This mimics the behaviour of Django's `ManyRelatedManager`

Take `test_returns_new_data_after_update`. It emulates the behaviour of Django-rest-framework's UpdateModelMixin.

If I did any `prefetch_related` on my QuerySet, the resulting data still uses the prefetched results instead of the updated data.

### DIFF
--- a/taggit/managers.py
+++ b/taggit/managers.py
@@ -143,10 +143,9 @@ class _TaggableManager(models.Manager):
         return self.through.lookup_kwargs(self.instance)
 
     def _remove_prefetched_objects(self):
-        try:
-            self.instance._prefetched_objects_cache.pop(self.prefetch_cache_name)
-        except (AttributeError, KeyError):
-            pass  # nothing to clear from cache
+        prefetch_cache = getattr(self.instance, "_prefetched_objects_cache", None)
+        if prefetch_cache:
+            prefetch_cache.pop(self.prefetch_cache_name, None)
 
     @require_instance_manager
     def add(self, *tags, through_defaults=None, tag_kwargs=None, **kwargs):
--- a/taggit/managers.py
+++ b/taggit/managers.py
@@ -142,9 +142,15 @@ class _TaggableManager(models.Manager):
     def _lookup_kwargs(self):
         return self.through.lookup_kwargs(self.instance)
 
+    def _remove_prefetched_objects(self):
+        try:
+            self.instance._prefetched_objects_cache.pop(self.prefetch_cache_name)
+        except (AttributeError, KeyError):
+            pass  # nothing to clear from cache
+
     @require_instance_manager
     def add(self, *tags, through_defaults=None, tag_kwargs=None, **kwargs):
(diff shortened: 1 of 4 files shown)


tests/test_serializers.py
```diff
@@ -72,3 +71,19 @@ def test_taggit_removes_tags(self):
...
         assert TestModel.objects.filter(tags__name__in=["1"]).count() == 0
         assert TestModel.objects.filter(tags__name__in=["1", "2"]).count() == 1
+
+    def test_returns_new_data_after_update(self):
+        """
+        Test if the the prefetch cache clears after an updates
+        """
+        TestModel.objects.create().tags.add("1")
+
+        test_model = TestModel.objects.prefetch_related("tags").get()
```
rtpg, line 81: Reading this test at first it wasn't clear what we were testing. 

I think it would be clearer if at first we add an assert here to show that the serializer data is `["1"]`. (`assert TestModelSerializer(test_model).data["tags"] == ["1"]`).

Then at the end of this we can do `set(serializer.data["tags"]) == {"2", "3"}`.

taggit/managers.py
```diff
@@ -142,9 +142,15 @@ def uuid_rel_obj_attr(v):
     def _lookup_kwargs(self):
         return self.through.lookup_kwargs(self.instance)
 
+    def _remove_prefetched_objects(self):
+        try:
+            self.instance._prefetched_objects_cache.pop(self.prefetch_cache_name)
+        except (AttributeError, KeyError):
+            pass  # nothing to clear from cache
```
rtpg, line 149: stylewise, I think it would be nice if this was structured more like:

```
obj_cache = getattr(self.instance, "_prefetched_objects_cache", None)
if obj_cache:
    obj_cache.pop(self.prefetch_cache_name, None)
```

That way we don't have a `try`/`except` catching `AttributeError` and `KeyError` (both pretty heavily used exceptions). 

I don't think the code here is wrong but I think my proposed alternative makes it clearer what the expected situations are (for example I don't think we are expecting `self.prefetch_cache_name` to through an `AttributeError`)

rtpg:
Wow, this is a very nasty bug! Thanks for the find! 

I'm not going to merge this just yet, I want to first confirm that the tests do fail without your change and will look at what Django does to see if we have any way of piggy-backing off of this.

 I left some comments, but it's mostly style comments and nothing blocking on a merge.

rtpg:
Please ping me directly if we are May 6th or beyond and I haven't added any updates here

rtpg:
Alrighty, I ran this locally, confirmed that the added tests fail without the change.

This is unfortunately a bug that exists because we are _not_ using the `RelatedManager` right now. That being said, this fix matches what Django does, and fixes an existing issue in a way that is satisfying to me. This is a squashed bug!

rtpg (approved):

PR #816, opened 2022-08-09
Add pyupgrade and django-upgrade pre-commmit hook

hramezani:


rtpg:
Not a super fan of this kind of tooling but we'll put it in and see how things goes

Issue #828, opened 2022-10-27
`TagListSerializerField` is not a `ListSerializer`


janrito:
drf-spectacular fails to understand that the field should be serialised into a list of strings, rather than a string.

@ollie-zhang submitted the issue to the the original [repo](https://github.com/glemmaPaul/django-taggit-serializer/) here https://github.com/glemmaPaul/django-taggit-serializer/issues/47

I botched a fix 

```
from drf_spectacular.extensions import OpenApiSerializerFieldExtension
from drf_spectacular.openapi import AutoSchema
from drf_spectacular.plumbing import build_array_type, build_basic_type  # type: ignore
from drf_spectacular.types import OpenApiTypes
from drf_spectacular.utils import Direction


class TagListSerializerFieldFix(OpenApiSerializerFieldExtension):
    target_class = "taggit.serializers.TagListSerializerField"

    def map_serializer_field(self, auto_schema: AutoSchema, direction: Direction):

        return build_array_type(build_basic_type(OpenApiTypes.STR))  # type: ignore
```

But it would be nice if it was not needed

rtpg:
Thank you for reporting this! Definitely sounds better if we use a `ListField` given... well.. it's a tag list! Thank you for posting the workaround as well.

PR #832, opened 2022-11-14
Update TagListSerializerField to ListField subclass

brittandeyoung:
# Description
This pull request is to resolve an issue with the current serializer. When using drf_yasg or other api documentation tools, the tags filed shows as a string instead of a list. This Pull request will update `TagListSerializerField` to properly inherit from the `ListField` class. 

Closes: #828

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -8,6 +8,7 @@ Changelog
   to stale reads in cases where `prefetch_related` is used.
 * Add Python 3.11 support.
 * Add Django 4.1 support.
+* Fixed an issue where TagListSerializerField caused the tags field to show a string.
 
 3.0.0 (2022-05-02)
 ~~~~~~~~~~~~~~~~~~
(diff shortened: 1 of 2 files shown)


brittandeyoung:
The Contributors guide states to also perform a version bump within the `taggit/__init__.py` file, but I am assuming a Maintainer does that when they release a new version? Just let me know and I can bump it if needed.

rtpg:
Alright, I'm looking at this branch and trying to determine a good test here that could catch this and others

rtpg:
@brittandeyoung I made #833, including some extra changes. Could you go over and look at that and review? 

For releases, like you inferred that happens during releases, which I do separately. Given that we have a bugfix along with this change, once this change is merged I'll prepare a release.

brittandeyoung:
@rtpg Yeah this looks great! I am already using a custom class with the listfield in my app and it is working great. Will be nice to be able to use this instead. 

Should I remove my pull request since you are handling it in another?

rtpg:
@brittandeyoung thanks for the check! I merged the changes over there, and will prepare a release. Thank you for the report and the report of success in production!

PR #833, opened 2022-11-14
Make TagListSerializerField subclass ListField

### LINKED ISSUE #828: `TagListSerializerField` is not a `ListSerializer`

rtpg:
This is a very tricky change, and I expect some issues once people begin using this in earnest. But it is closer to where we want to be, and the main functionality is being tested. 

This change is motivated by documentation tools not properly identifying this field as a `ListField`.

This also includes ome documentation tweaks in the changelog

Thanks to @brittandeyoung for the initial PR

Fixes #828

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -5,9 +5,10 @@ Changelog
 ~~~~~~~~~~~~
 
 * Fixed an issue where object caches would not be properly cleared after updating tags, leading
-  to stale reads in cases where `prefetch_related` is used.
+  to stale reads in cases where ``prefetch_related`` is used.
 * Add Python 3.11 support.
 * Add Django 4.1 support.
+* Change ``TagListSerializerField`` to be a subclass of ``ListField``. This should improve support for API document generation. This change should not affect API behavior, but might affect metaprogramming code, so please procede carefully during this update.
 
 3.0.0 (2022-05-02)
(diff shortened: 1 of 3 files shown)


brittandeyoung:
@rtpg Yeah this looks great! I am already using a custom class with the listfield in my app and it is working great. Will be nice to be able to use this instead.

rtpg:
the codecov reprot incorrectly reported a coverage drop, but that went back to 0% change.

PR #850, opened 2023-04-13
fix(tests): Add contains_over_clause #848

imAsparky:
Added support for Django 4.2

Adding support for Django 4.2 broke some tests with this error `AttributeError: 'ExtraJoinRestriction' object has no attribute 'contains_over_clause'`

Tracking the source of the error I have found a change to `38  def split_having(self, negated=False):` in  [django 4.1](https://github.com/django/django/blob/stable/4.1.x/django/db/models/sql/where.py) 
to  `38 def split_having_qualify(self, negated=False, must_group_by=False):` in [django 4.2](https://github.com/django/django/blob/stable/4.2.x/django/db/models/sql/where.py) which is causing the error.

The[ `'contains_over_clause'`](https://docs.djangoproject.com/en/4.2/ref/models/expressions/#expression-api)  Attribute Tells Django that this expression contains a [Window](https://docs.djangoproject.com/en/4.2/ref/models/expressions/#django.db.models.expressions.Window) expression. It’s used, for example, to disallow window function expressions in queries that modify data. 
Examples of what the OVER clause does can be seen in the [postgres documentation](https://www.postgresql.org/docs/current/tutorial-window.html#:~:text=The%20OVER%20clause%20determines%20exactly,PARTITION%20BY%20expression(s).).

Adding `contains_over_clause = False` to `ExtraJoinRestriction` allows the tests to run successfully.

closes #848

PR #847, opened 2023-04-03
Update GH Actions and tox.

imAsparky:
GitHub tests  workflow.

* Add a workaround for tests to pass, see issue #846 for info and  a link to GH issue/discussion.  Users have pinned to Ubuntu 20.04 until a solution is found. 

* Update actions/checkout@v2 to actions/checkout@v3 to silence deprecation warnings.

Tox Django test matrix.

* Updated dj-main to minimum py3.10.

closes #846

PR #852, opened 2023-04-20
Fix test failures against Django `main`

### LINKED ISSUE #851: Django 5.0 test failures
In addition to #848, we have some more test failures that are specific to Django 5.0.

- `AttributeError: 'TaggableManager' object has no attribute 'get_reverse_joining_fields'`
  `get_reverse_joining_columns` has been deprecated in favour of `get_reverse_joining_fields` as of https://github.com/django/django/pull/16632.
- Django defaults to div-based form rendering in 5.0
   - https://github.com/django/django/commit/98756c685ee173bbd43f21ed0553f808be835ce5
   - https://github.com/django/django/commit/232b60a21b951bd16b8c95b34fcbcbf3ecd89fca

laymonage:
Fixes #851 (at least for now).

This is built on top of #850, and with a different approach for #847.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,7 +5,8 @@ on: [push, pull_request]
 jobs:
   build:
     name: Python ${{ matrix.python-version }}
-    runs-on: ubuntu-latest
+    # https://github.com/actions/setup-python/issues/544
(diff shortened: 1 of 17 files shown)


rtpg:
Thank you for the patches, I pulled these changes in (I went for removing Python 3.6 support outright, as its at EOL)

PR #860, opened 2023-06-22
Fix `tag_kwargs` and case insensitive discrepancy

### LINKED ISSUE #859: `tag_kwargs` are not taken into account when `TAGGIT_CASE_INSENSITIVE` is `True`
`tag_kwargs` are not taken into account when case insensitivity is turned on, while they are when it is off.
For cases like #809 that would result in already existing tags, with the same name, from other users/tenants being used instead of creating a new one.

When case insensitivity is turned off, this will work as expected as the `tag_kwargs` are used in the filter to check for existing tags.

In `_TaggableManager._to_tag_model_instances`:
https://github.com/jazzband/django-taggit/blob/63c9a97066cdd3d0dc0ed1283eefacff54d605e2/taggit/managers.py#L221-L238

If I change the get call on line 229 to `get(name__iexact==name, **tag_kwargs)` it works the same for both cases.

dennisv:
Fixes #859


rtpg:
Thank you for catching this! This is clearly a rather big bug, I'll make sure to put out a release in the next couple of weeks to handle this

PR #862, opened 2023-07-06
Fix Django warnings about index_together with new migration

AlanCoding:
This is triggering a Django warning

> RemovedInDjango51Warning: 'index_together' is deprecated. Use 'Meta.indexes' in 'taggit.TaggedItem' instead.

This updates the `index_together` use to resolve this, hopefully painlessly.


rtpg:
something to check, whether index renaming is cheap or not in the major SQL vendors (locking etc)

rtpg:
also whether the index renaming is actually safe? we want to make sure that deployments including this don't cause downtime.

rtpg:
Things to do on this one:
-  set things up to make `RemovedInXWarning`s instead be errors for testing on `djmain`. 
-  concretise the existing index names when setting up the `IndexTogether` relations, making sure to not generate any actual SQL operations

mister-rao:
Getting this error while running tests:

```
django.utils.deprecation.RemovedInDjango51Warning: 'index_together' is deprecated. Use 'Meta.indexes' in 'taggit.TaggedItem' instead. 
``` 

Versions:
Django = "4.2.2"
django-taggit = "4.0.0"

igorsimb:
Why is this not merged yet?

AlanCoding:
Addressed in https://github.com/jazzband/django-taggit/pull/869

PR #871, opened 2023-10-11
Have tags default to being in creation order

### LINKED ISSUE #870: Tag Ordering Inconsistent
Looking at the docs, it appears that the `tags.all()` queryset should maintain the ordering of the tags as they were initially set. However, in testing this does not hold. The tags do not in fact maintain their original ordering. How can I retrieve the tags with their original ordering intact?

Please note the sample below which shows that the ordering is not maintained as the docs show.

```
>>> apple = Food.objects.create()
>>> apple
<Food: Food object (1)>
>>> apple.tags.add("red", "green", "delicious")
>>> apple.tags.all()
<QuerySet [<Tag: green>, <Tag: red>, <Tag: delicious>]>
```

steverecio:
Fixes #870 to ensure that tag order is preserved by default

### DIFF
--- a/taggit/managers.py
+++ b/taggit/managers.py
@@ -69,7 +69,9 @@ class _TaggableManager(models.Manager):
         if ordering:
             self.ordering = ordering
         else:
-            self.ordering = []
+            # default to ordering by the pk of the through table entry
+            related_name = self.through.tag.field.related_query_name()
+            self.ordering = [f"{related_name}__pk"]
 
     def is_cached(self, instance):
         return self.prefetch_cache_name in instance._prefetched_objects_cache
--- a/taggit/managers.py
+++ b/taggit/managers.py
@@ -221,14 +221,18 @@ class _TaggableManager(models.Manager):
                             tag_objs.append(tag)
                         except self.through.tag_model().DoesNotExist:
                             lookup = {"name__iexact": t, **tag_kwargs}
-                            tag, created = manager.get_or_create(**lookup, defaults={"name": t})
+                            tag, created = manager.get_or_create(
+                                **lookup, defaults={"name": t}
+                            )
                             tag_objs.append(tag)
                     else:
                         try:
                             tag = manager.get(name=t, **tag_kwargs)
                             tag_objs.append(tag)
                         except self.through.tag_model().DoesNotExist:
-                            tag, created = manager.get_or_create(name=t, defaults={"name": t})
+                            tag, created = manager.get_or_create(
+                                name=t, defaults={"name": t}
+                            )
                             tag_objs.append(tag)
             else:
                 raise ValueError(
@@ -238,7 +242,6 @@ class _TaggableManager(models.Manager):
                 )
 
         return tag_objs
-
 
     @require_instance_manager
     def names(self):
--- a/taggit/managers.py
+++ b/taggit/managers.py
@@ -196,18 +196,38 @@ class _TaggableManager(models.Manager):
     def _to_tag_model_instances(self, tags, tag_kwargs):
         """
         Takes an iterable containing either strings, tag objects, or a mixture
-        of both and returns set of tag objects.
+        of both and returns a list of tag objects while preserving order.
         """
         db = router.db_for_write(self.through, instance=self.instance)
 
-        str_tags = set()
-        tag_objs = set()
+        processed_tags = []
+        tag_objs = []
+
+        case_insensitive = getattr(settings, "TAGGIT_CASE_INSENSITIVE", False)
+        manager = self.through.tag_model()._default_manager.using(db)
 
         for t in tags:
             if isinstance(t, self.through.tag_model()):
-                tag_objs.add(t)
+                if t not in tag_objs:
+                    tag_objs.append(t)
             elif isinstance(t, str):
-                str_tags.add(t)
+                if t not in processed_tags:
+                    processed_tags.append(t)
+                    if case_insensitive:
+                        try:
+                            tag = manager.get(name__iexact=t, **tag_kwargs)
+                            tag_objs.append(tag)
+                        except self.through.tag_model().DoesNotExist:
+                            lookup = {"name__iexact": t, **tag_kwargs}
+                            tag, created = manager.get_or_create(**lookup, defaults={"name": t})
+                            tag_objs.append(tag)
+                    else:
+                        try:
+                            tag = manager.get(name=t, **tag_kwargs)
+                            tag_objs.append(tag)
+                        except self.through.tag_model().DoesNotExist:
+                            tag, created = manager.get_or_create(name=t, defaults={"name": t})
+                            tag_objs.append(tag)
             else:
                 raise ValueError(
                     "Cannot add {} ({}). Expected {} or str.".format(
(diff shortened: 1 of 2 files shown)


taggit/managers.py
```diff
@@ -196,58 +198,49 @@ def add(self, *tags, through_defaults=None, tag_kwargs=None, **kwargs):
...
+                                **lookup, defaults={"name": t}
+                            )
+                            tag_objs.append(tag)
+                    else:
+                        try:
+                            tag = manager.get(name=t, **tag_kwargs)
+                            tag_objs.append(tag)
+                        except self.through.tag_model().DoesNotExist:
+                            tag, created = manager.get_or_create(
+                                name=t, defaults={"name": t}
```
rtpg, line 234: I think you're missing `**tag_kwargs` here

taggit/managers.py
```diff
@@ -69,7 +69,9 @@ def __init__(self, through, model, instance, prefetch_cache_name, ordering=None)
         if ordering:
```
rtpg, line 69: if we're going to set a default for `self.ordering`, let's make it so that we can still pass in `ordering=[]`, by  changing this `if ordering` to `if ordering is not None`.

That way in the changelog we can offer people to have the old behavior by passing that in explicitly

rtpg (changes requested):
When reading the original issue I was a bit pessimistic, but looking at this I think this makes more sense than I was originally expecting.

The test failures seem to be mostly linked to `tag_kwargs` not being passed into `get_or_create`, so we should do that IMO.

The point about allowing `self.ordering = []` to be explicitly set is also important IMO: sometimes people don't want stuff ordered implicitly (can be the source of nasty DB plans)

taggit/managers.py
```diff
@@ -196,58 +198,49 @@ def add(self, *tags, through_defaults=None, tag_kwargs=None, **kwargs):
     def _to_tag_model_instances(self, tags, tag_kwargs):
         """
         Takes an iterable containing either strings, tag objects, or a mixture
-        of both and returns set of tag objects.
+        of both and returns a list of tag objects while preserving order.
```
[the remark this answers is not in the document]
steverecio: > So here this preserves the order of creation, right? But if someone has ordering set to name should this actually order by name instead?

I would argue that name (or tag creation order) is much easier to order by on the retrieval end. By preserving the order of the original list in this function, we can order by any key we want when retrieving the model (tag name, tag pk, or through table entry pk).

tests/test_models.py
```diff
@@ -1,8 +1,39 @@
 from django.test import TestCase, override_settings
 
+from taggit.models import Tag
 from tests.models import TestModel
 
 
+class TestTaggableManager(TestCase):
```
steverecio, line 7: Not sure if this suite of tests should be in this file or tests.py

steverecio:
So as of now the primary tests that are failing are the [assertNumQueries](https://github.com/jazzband/django-taggit/blob/master/tests/tests.py#L523) and [here](https://github.com/jazzband/django-taggit/blob/master/tests/tests.py#L194) which seems to occur because the ordering query causes duplicates but I tried adding `distinct()` in everywhere I could but that doesn't seem to do it. I think its somewhere in the caching logic but can't quite figure it out.

@rtpg Are we able to modify the tests to change the num_queries from 32 to 34 and 10 to 11?

steverecio:
Continuing to dig into [this failure](https://github.com/jazzband/django-taggit/blob/master/tests/tests.py#L194)...

The issue appears to come from [`get_queryset`](https://github.com/jazzband/django-taggit/blob/master/taggit/managers.py#L76). If I print the query getting executed, there is already a distinct clause but because of the through table entry on the multi-inheritance model the distinct clause is ineffective in deduplicating the query. However, we can't pass in the tag pk for deduplicating because we would have to pass in the tag pk to the order_by query for distinct() to accept that param.

Here is the query:

```
SELECT DISTINCT "taggit_tag"."id", "taggit_tag"."name", "taggit_tag"."slug", "taggit_taggeditem"."id" FROM "taggit_tag" INNER JOIN "taggit_taggeditem" ON ("taggit_tag"."id" = "taggit_taggeditem"."tag_id") INNER JOIN "django_content_type" ON ("taggit_taggeditem"."content_type_id" = "django_content_type"."id") WHERE ("django_content_type"."app_label" = tests AND "django_content_type"."model" = food AND "taggit_taggeditem"."object_id" = 1) ORDER BY "taggit_taggeditem"."id" ASC
```

rtpg:
>  SELECT DISTINCT "taggit_tag"."id", "tests_taggedcustompk"."id" FROM "taggit_tag" INNER JOIN "tests_taggedcustompk" ON ("taggit_tag"."id" = "tests_taggedcustompk"."tag_id") INNER JOIN "django_content_type" ON ("tests_taggedcustompk"."content_type_id" = "django_content_type"."id") WHERE ("django_content_type"."app_label" = 'tests' AND "django_content_type"."model" = 'custompkfood') ORDER BY "tests_taggedcustompk"."id" ASC

On `test_most_common_lazy`, this query is getting generated. This looks like something pretty hefty (my first read is that this is going to read _all_ of the tags of a certain type in from the database). I think we will want to figure out a way to avoid this (and somehow make this extra query lazy again, as I bet in that case the ORM will do the right thing).

 I scrolled through the diff on the web just now and can't figure out exactly what line introduced this query though.

steverecio:
> > SELECT DISTINCT "taggit_tag"."id", "tests_taggedcustompk"."id" FROM "taggit_tag" INNER JOIN "tests_taggedcustompk" ON ("taggit_tag"."id" = "tests_taggedcustompk"."tag_id") INNER JOIN "django_content_type" ON ("tests_taggedcustompk"."content_type_id" = "django_content_type"."id") WHERE ("django_content_type"."app_label" = 'tests' AND "django_content_type"."model" = 'custompkfood') ORDER BY "tests_taggedcustompk"."id" ASC
> 
> On `test_most_common_lazy`, this query is getting generated. This looks like something pretty hefty (my first read is that this is going to read _all_ of the tags of a certain type in from the database). I think we will want to figure out a way to avoid this (and somehow make this extra query lazy again, as I bet in that case the ORM will do the right thing).
> 
> I scrolled through the diff on the web just now and can't figure out exactly what line introduced this query though.

This is related to the new `get_queryset` function: https://github.com/jazzband/django-taggit/pull/871/files#diff-2f4b79363da4c4d9048dbf680784ffbdee2400138f9672c19e2e7e73e59093d0R89

I had to execute the query in order to dedupe it because the `distinct()` function doesn't seem to work with the ordering query attached. Not sure how to get the distinct + ordering working together...

steverecio:
@rtpg friendly ping :)

I was able to fix most of the tests by only applying the ordering if the instance is not None. This way, the ordering is not applied automatically when querying for tags attached to a model, only for tags attached to a model instance. Only remaining issue is the 34 vs 32 queries.

stefanivic:
@steverecio @rtpg any updates on this, we ran into the same issue so was wondering if this feature was abandoned or will not be merged due to some other reason ? I can take it over if you moved on.

rtpg:
@stefanivic could you specifically describe the issue you had, so we can design some test cases and properly say what issue we're trying to fix?

Basically, would like to fix any issues, but do not want to add any extra queries, so I think we might need to start a new branch. But would like to have a specific ask here

stefanivic:
> @stefanivic could you specifically describe the issue you had, so we can design some test cases and properly say what issue we're trying to fix?
> 
> Basically, would like to fix any issues, but do not want to add any extra queries, so I think we might need to start a new branch. But would like to have a specific ask here

Pretty much what the referenced issue and PR talk about. When adding tags, their order is not respected, instead the order seems to change during the save. For example if I add the tags `1, 3, 5, 2, 4`, after the save these will be listed as `1, 2, 3, 4, 5`. Based on what I have seen it seems to order them by name ( alphabetical asc ). If it's required I can make a small demo to show. 

Our clients tend to define tags in the order of importance to them, so the ones they want people to click the most go on the first two positions. But after the reorder they can end up last.

I have tried adding a sort_order (int) field to the ItemBase and then returning them in the specific order, but the Set() seems to reorder them regardless.

steverecio:
> @stefanivic could you specifically describe the issue you had, so we can design some test cases and properly say what issue we're trying to fix?
> 
> Basically, would like to fix any issues, but do not want to add any extra queries, so I think we might need to start a new branch. But would like to have a specific ask here

@rtpg There are some new [test cases](https://github.com/jazzband/django-taggit/pull/871/files#diff-0adb667cd397bd56edce10eec11e1b10821740a10d72bd148b09645fc9108968) in this PR that highlight the issue.

rtpg:
This one was implemented in #892

PR #876, opened 2023-10-25
Update minimal dependency version of Django to 4.1

### LINKED ISSUE #875: Django Taggit 5.0 is not compatible with Django <4.1
The latest [0006 migration][1] uses the `migrations.RenameIndex` method which has been [introduced in Django v4.1][2]. Running it on Django 3.2 will result in error below:

> AttributeError: module 'django.db.migrations' has no attribute 'RenameIndex'. Did you mean: 'RemoveIndex'?

The minimum Django version [here][3] should be raised accordingly to prevent installing it automatically.

[1]: https://github.com/jazzband/django-taggit/blob/master/taggit/migrations/0006_rename_taggeditem_content_type_object_id_taggit_tagg_content_8fc721_idx.py
[2]: https://docs.djangoproject.com/en/4.2/ref/migration-operations/#renameindex
[3]: https://github.com/jazzband/django-taggit/blob/master/setup.cfg#L37

bartTC:
This is in line with the current test matrix. Fixes #875

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -34,7 +34,7 @@ project_urls =
 [options]
 python_requires = >=3.6
 packages = find:
-install_requires = Django>=3.2
+install_requires = Django>=4.1
 include_package_data = true
 zip_safe = false
 


rtpg:
@bartTC thanks for sending this in and signaling the original issue. I just released 5.0.1, which includes this fix as well as some more fixes (and a note in the release notes to make sure to not miss this in the future)

PR #877, opened 2023-10-26
Prepare v5.0.1

rtpg:
This just tags the requirements properly.

@jezdez I think we want to yank 5.0.0 (after I release 5.0.!), just because Django 3.2 is an LTS release so until mid-2024. Is there a process that would allow me to do yanking on Pypi?

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,12 @@ Changelog
 (Unreleased)
 ~~~~~~~~~~~~
 
+5.0.1 (2023-10-26)
+~~~~~~~~~~~~~~~~~~
+
+* Fix the package metadata to properly reflect the right Django and Python version requirements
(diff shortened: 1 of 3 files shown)


jezdez:
@rtpg No, yanking isn't implemented on the Jazzband side since it predates the PyPI feature, but I can do that manually.

rtpg:
Alright, 5.0.1 (which fixes the constraints)

@jezdez would you be able to yank the 5.0.0 release? My understanding of yanking is that people who have pinned to 5.0.0 will still have access to it (fine!) but people who install without pinning will not have it proposed to them (good, because 5.0.0's django requirements state 3.2 support when it's in fact not supporting 3.2). So people who want 5.0.0 can still have it, and people on Django 3.2 will not be offered an incompatible release anymore.

jezdez:
@rtpg Done

PR #886, opened 2024-01-30
Update Chinese translation

zengqiu:
Update Chinese translation (`locale/zh_Hans`).


rtpg:
Thank you for sending these in! In the future it would be helpful to avoid including other changes with the translations but I think that `.gitignore` change is fine :+1:

Issue #899, opened 2024-06-23
Offer Tag Merging


rtpg:
Basically there are times where you want to say "I have Tag A and Tag B, let's have only Tag A". This would involve rewriting any `TagItem`s and friends that are using Tag B to instead point to Tag A (and at the very end probably delete Tag B?)

Ideas of what we could offer:

- a Python function like `tag_b.merge_into(tag_a)` (this would make tag A the canonical one)
- a Django management command that could be used to do this merging
- Something in the Admin where we could click on a tag, then say to merge that tag into another one

I _think_ there's gotta be some functions like "find django objects that refer to this object", given that the Django Admin lists related items on delete...

rtpg:
(#470 is related to this)

Trafire:
Was using this as a way to explore the library this morning, Maybe something like this would work?

```python
from taggit.models import Tag, TaggedItem


def update_vs_delete_tags(base_tag_name):
    """
    Identifies tags that are duplicates of the base_tag_name case insensitive value.
    
    Ignores any duplicate values.
        
    """
    duplicate_tags = Tag.objects.filter(name__iexact=base_tag_name).exclude(tag__name=base_tag_name)
    existing_items = set(TaggedItem.objects.filter(tag__name=base_tag_name).values_list("object_id", "content_type_id"))
    to_update = []
    tagged_items = TaggedItem.objects.filter(tag__in=duplicate_tags)
    for tagged_item in tagged_items:
        comparison = (tagged_item.object_id, tagged_item.content_type_id)
        if comparison not in existing_items:
            to_update.append(tagged_item.id)
        existing_items.add(comparison)
    return to_update


def merge_case_insensitive_tags(base_tag_name):
    to_update = update_vs_delete_tags(base_tag_name)
    base_tag, _ = Tag.objects.get_or_create(name=base_tag_name)
    TaggedItem.objects.filter(tag__in=to_update).update(
        tag_id=base_tag.id
    )
    # Remove All Tags that are now redundant
    Tag.objects.filter(name__iexact=base_tag_name).exclude(tag__name=base_tag_name)
    
```

PR #904, opened 2024-07-14
Enable Tag Merging

guel-codes:
## Overview
This pertains to this issue #899

## Updates:
- `CHANGELOG.rst`
- `docs/admin.rst`
- `taggit/admin.py`
- `taggit/forms.py`
- `taggit/templates/admin/taggit/merge_tags_form.html`

## Demo
https://www.loom.com/share/72db6fb95c7b4b4dab90746d6ad2fd32?sid=dbda0b6a-ee58-4f93-b2bd-67f677a554e9


taggit/admin.py
```diff
@@ -14,3 +19,66 @@ class TagAdmin(admin.ModelAdmin):
...
+    def render_tag_form(self, request, queryset):
+        selected = request.POST.getlist(admin.helpers.ACTION_CHECKBOX_NAME)
+        if not selected:
+            self.message_user(request, "Please select at least one tag.")
+            return redirect(request.get_full_path())
+
+        selected_tag_ids = ",".join(selected)
+        redirect_url = (
+            f"{request.get_full_path()}merge-tags/?selected_tags={selected_tag_ids}"
+        )
```
rtpg, line 45: instead of sending them over as GET params, I guess we could just from this point store the tag IDs in the session at this point and redirect over to the `merge_tags_view`? That way we avoid a serialization roundtrip (incidentally also avoid issues with GET param length as well)

taggit/admin.py
```diff
@@ -14,3 +19,66 @@ class TagAdmin(admin.ModelAdmin):
...
+                        tag = Tag.objects.get(id=tag_id)
+                        tagged_items = TaggedItem.objects.filter(tag=tag)
+                        for tagged_item in tagged_items:
+                            tagged_item.tag = new_tag
+                            tagged_item.save()
+                        # tag.delete()  #this will delete the selected tags after merge
+
+                # clear the selected_tag_ids from session after merge is complete
+                request.session.pop("selected_tag_ids", None)
+                return redirect("..")
```
rtpg, line 74: Does that work? TIL

If that doesn't work and you want to always redirect to the admin list view, there is some name given by the admin (I think it's `admin_model_list_view`? With model names as params).

rtpg: Also, I think it would be worth calling `self.message_user` with a success message when we successfully "do the thing", so to speak.

taggit/admin.py
```diff
@@ -14,3 +19,66 @@ class TagAdmin(admin.ModelAdmin):
...
+            if form.is_valid():
+                new_tag_name = form.cleaned_data["new_tag_name"]
+                new_tag, created = Tag.objects.get_or_create(name=new_tag_name)
+                with transaction.atomic():
+                    for tag_id in selected_tag_ids:
+                        tag = Tag.objects.get(id=tag_id)
+                        tagged_items = TaggedItem.objects.filter(tag=tag)
+                        for tagged_item in tagged_items:
+                            tagged_item.tag = new_tag
+                            tagged_item.save()
```
rtpg, line 69: I think that here, we probably should go for an `update` call, even though it won't call `.save`. Something like `TaggedItem.objects.filter(tag=tag).update(tag=new_tag)`. That way we will support tags that already have loads of tagged items.

taggit/admin.py
guel-codes: The redirect does work. Since I am appending the `merge-tags/` during the first redirect to the form,  this goes back one level to the path before which is the list of tags.

guel-codes: I added the `self.message_user` to notify the user after the merge was complete but I think the redirect back to the previous page is happen too quickly to see it. Maybe we can find a better way to notify the users.

taggit/admin.py
guel-codes: Yeah I think this is cleaner, updating now

taggit/admin.py
rtpg: Hmmm... normally the mesasge should appear on the "next" page load so to speak (if this is happening through the Django `messages` framework), will take a look.

rtpg: Try `self.message_user(request, "some message", messages.SUCCESS)`, that should work (I looked inside the definition of `ModelAdmin` to find some examples)

taggit/admin.py
```diff
@@ -14,3 +19,60 @@ class TagAdmin(admin.ModelAdmin):
...
+                            TaggedItem.objects.filter(tag=tag).update(tag=new_tag)
+                        # tag.delete()  #this will delete the selected tags after merge...leaving out for now
+
+                self.message_user(request, "Tags have been merged")
+                # clear the selected_tag_ids from session after merge is complete
+                request.session.pop("selected_tag_ids", None)
+                return redirect("..")
+            else:
+                print(f"Form errors: {form.errors}")
+                self.message_user(request, "Form is invalid.", level="error")
```
rtpg, line 71: I think here you want to do `level=messages.ERROR` (they're numbers internally, not strings)

taggit/admin.py
```diff
@@ -14,3 +19,60 @@ class TagAdmin(admin.ModelAdmin):
...
+
+        context = {
+            "form": MergeTagsForm(),
```
rtpg, line 74: So when you're validating a form, you want to make sure you pass that form you tried to validate back into the context.

```
form = MergeTagsForm(data=request.POST  if request.method == "POST" else None)
if request.method == "POST":
  if form.is_valid():
   ....
...
context = {
  "form": form, 
   ...
}
```
if you do that the form errors will be properly displayed if there are any

taggit/forms.py
```diff
@@ -52,3 +52,11 @@ def has_changed(self, initial_value, data_value):
         initial_value.sort()
 
         return initial_value != data_value
+
+
+class MergeTagsForm(forms.Form):
+    new_tag_name = forms.CharField(
+        label="New Tag Name",
+        max_length=255,
```
rtpg, line 60: if you want to test your validation, protip: set `max_length` to `2` temporarily and you should be able to more easily see errors.

taggit/templates/admin/taggit/merge_tags_form.html
```diff
@@ -0,0 +1,29 @@
+<div id="mergeTagsModal">
```
rtpg, line 1: Maybe `{% extends "admin/base_site.html" %}`, then surround all of this with a `{% block content %}/ {% endblock %}`, and the form will look a _bit_ more like an admin page

rtpg: actually maybe just `admin/base.html`.

taggit/admin.py
```diff
@@ -14,3 +19,60 @@ class TagAdmin(admin.ModelAdmin):
...
+                            tagged_item.tag = new_tag
+                            TaggedItem.objects.filter(tag=tag).update(tag=new_tag)
+                        # tag.delete()  #this will delete the selected tags after merge...leaving out for now
+
+                self.message_user(request, "Tags have been merged")
+                # clear the selected_tag_ids from session after merge is complete
+                request.session.pop("selected_tag_ids", None)
+                return redirect("..")
+            else:
+                print(f"Form errors: {form.errors}")
```
rtpg, line 70: let's make sure to remove this before merging things in

taggit/admin.py
guel-codes: the `messages.SUCCESS` didn't work here but `level="success` got it working

taggit/forms.py
guel-codes: I tested this earlier and the max_length works, but it shouldn't throw an error at all because the form input box will be be limited to the max_length. For instance when I set it to 2 and try to add a tag "fruit" it will stop at "fr" and not let me input any other characters after that.

taggit/templates/admin/taggit/merge_tags_form.html
guel-codes: Updated this and also changed the button to an input to make the styling a little cleaner

taggit/admin.py
guel-codes: I ran into this above but the `mesages.ERROR` didn't work here either only the `level="error"`

taggit/admin.py
guel-codes: This isn't working because the context is defined outside of the conditional where the form is defined. I assumed the form that got passed in from `MergeTagsForm()` would work. It renders the form and that's how the max_length validation is implemented.

rtpg:
Just noticed this one doesn't have a test, will add one tomorrow.

Issue #708, opened 2021-02-01
Natural key support


lucemia:
Looks like `taggit` doesn't yet support loaddata / dumpdata with natural keys. 
https://docs.djangoproject.com/en/3.1/topics/serialization/#natural-keys

It will be great if django-taggit can add this feature.

rtpg:
@lucemia here I guess you're suggesting that we just use a straightforward mapping of tag text to do a lookup? Seems pretty doable (with the usual caveats for custom tag types)

Trafire:
Draft language for docs 
## Natural Keys Support

We have added [natural keys](https://docs.djangoproject.com/en/5.0/topics/serialization/#natural-keys) support to the custom Tag model in the Django taggit library. This allows you to identify objects by human-readable identifiers rather than by their database ID. 

```shell
python manage.py dumpdata taggit.Tag --natural-foreign --natural-primary
```

By default tags use the name field as the natural key. 

You can customize this in your own custom tag model by setting the `natural_key_fields` property on your model the required fields.

PR #905, opened 2024-07-14
708 - Add Natural Key Support to Tag model.

Trafire:
#708 

Adds the ability to load/dump Tags by by natural keys rather thank PK


taggit/models.py
```diff
@@ -15,7 +17,29 @@ def unidecode(tag):
         return tag
 
 
-class TagBase(models.Model):
+class NaturalKeyManager(models.Manager):
```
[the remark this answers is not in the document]
Trafire: Yeah, definitely want to do something here with organization. Moving this to managers.py  was my initial thought, but that creates a circular import with models.py. So do need to think a little on what the elegant solution is here.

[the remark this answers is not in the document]
guel-codes: Yeah, that's a good call. Maybe we leave it here for now.

taggit/models.py
```diff
@@ -26,6 +50,9 @@ class TagBase(models.Model):
         allow_unicode=True,
     )
 
+    natural_key_fields = ["name"]
+    objects = NaturalKeyManager(natural_key_fields)
```
rtpg, line 54: I think it should be possible for `NaturalKeyManager` to look up the keys through the `Model` it is managing (probably `self.model` in the context of a manager?)

If we did that then we could just have:
```
natural_key_fields = ["name"]
objects = NaturalKeyManager()
```

taggit/models.py
```diff
@@ -15,7 +17,29 @@ def unidecode(tag):
         return tag
 
 
-class TagBase(models.Model):
+class NaturalKeyManager(models.Manager):
+    def __init__(self, natural_key_fields: List[str], *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.natural_key_fields = natural_key_fields
```
rtpg, line 23: oh I think we're not actually using this? Since we call `self.model.natural_key_fields` anyways. Maybe I'm missing something

taggit/models.py
Trafire: Yep makes sense, let me clean this up

tim-schilling:
@rtpg when you have time, can you apply the djangonauts-reserved label to the following PRs (github allows you to mass apply labels from the list view):

- https://github.com/jazzband/django-taggit/pulls?q=is%3Apr+author%3ATrafire
- https://github.com/jazzband/django-taggit/pulls?q=is%3Apr+author%3Aguel-codes

PR #914, opened 2024-08-01
[#826] - get() returned more than one Tag -- it returned 2!

### LINKED ISSUE #826: get() returned more than one Tag -- it returned 2!
Hello
Sometimes when adding tags to an existing object, I am getting the following error.
```
get() returned more than one Tag -- it returned 2!
```
Here is the code I am using
```python3
...
publication.tags.set(all_tags, clear=True)
publication.save()
# I also tried
publication.tags.set(all_tags)
publication.save()
# and
publication.tags.add(*all_tags)
publication.save()
# but still getting the same issue
...
```
**Notes**:
- I am not using any get function
- Django version 4.1

fazeelghafoor:
Fixes #826 

- add check for MultipleObjectsReturned exception returning only the first tag when get() returns more than one
- add deduplicate_tags management command which keep the first tag and merge all duplicates into it (relationships), then delete the duplicates.
- add test for deduplicate_tags command


fazeelghafoor:
@rtpg please review and let me know if any changes required.

rtpg:
Thanks for the PR! I made some changes to simplify the management command and make it explicit that the ordering is always by `pk` (though that's a bit spurious given the behavior of `first`).

PR #917, opened 2024-08-15
Add an admin action to remove orphaned tags

guel-codes:
## Overview
We have a management command to remove orphaned tags but I also wanted to add functionality to do it as an admin action from the drop down list.

## Preview
- https://www.loom.com/share/3cd37182c652493997bbfd92f3eab5f9?sid=78dc947a-102c-4f53-800e-368264ac039f

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,8 @@ Changelog
 (Unreleased)
 ~~~~~~~~~~~~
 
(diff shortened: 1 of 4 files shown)


rtpg:
Alright, tweaked this one up to have an `orphaned` queryset method, and so that the orphan cleanup only happens within the selected tags (matching the principle that actions occur only among the selected data, it's two clicks to select everything)

Issue #900, opened 2024-06-24
[October 2024] Deprecated Support of Python 3.8


guel-codes:
Python 3.8 will reach EoL status in 10/2024. Once that happens we should remove support for Python 3.8 and update our testing suite.
https://devguide.python.org/versions/

guel-codes:
Hey @rtpg, just wanted to remind you about this one. I saw 3.13 was released a couple days ago and 3.8 has officially been moved into EoL status. I will try to get a PR up for this for you to review sometime this week.

PR #925, opened 2024-10-18
Removing 3.8

guel-codes:
Overview:
This PR is mainly for cleanup. Now that Python 3.8 has officially reached EoL, it is time to remove support for that version.
This refers to #900

Updates:
These 2 are probably the most important for review since they handle CI/CD workflows
- `.github/workflows/release.yml`
- `.github/workflows/test.yml`

########################

- `CHANGELOG.rst`
- `docs/index.rst`
- `setup.cfg`
- `taggit/__init__.py`
- `tox.ini`

### DIFF
--- a/taggit/__init__.py
+++ b/taggit/__init__.py
@@ -1,2 +1,2 @@
-VERSION = (6, 1, 1)
+VERSION = (6, 1, 0)
 __version__ = ".".join(str(i) for i in VERSION)
--- a/taggit/__init__.py
+++ b/taggit/__init__.py
@@ -1,2 +1,2 @@
-VERSION = (6, 1, 0)
+VERSION = (6, 1, 1)
 __version__ = ".".join(str(i) for i in VERSION)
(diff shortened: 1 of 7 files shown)


taggit/__init__.py
```diff
@@ -1,2 +1,2 @@
-VERSION = (6, 1, 0)
```
rtpg, line 1: Let's avoid bumping this up for now, we'll bump it up once we make a release

tox.ini
```diff
@@ -4,15 +4,14 @@ envlist =
...
+    py{39,310,311,312}-dj{41,42}
     py{310,311,312}-dj{50}
     py{310,311,312}-djmain
     docs
 
 [gh-actions]
 python =
-    3.8: py38, black, flake8, isort
-    3.9: py39
+    3.9: py39, black, flake8, isort
```
rtpg, line 14: Glad to see we didn't lose these, that was one thing I was worried about

taggit/__init__.py
guel-codes: Sounds good 🙌🏽

PR #927, opened 2024-11-02
Fix bug with merge form action when querystring is present.

garrettc:
Querystrings are preventing the tags from merging, because `get_full_path()` in the redirect preserves querystrings.

**Steps to reproduce**

1. Load the admin area > Taggit > Tags
2. Sort by `slug` so that the path is `/admin/taggit/tag/?o=2.1` (with querystring)
3. Select two tags, eg: "Adventure" and "Anthology, then action "Merge selected tags"

**Result**

Because the resulting path is `/admin/taggit/tag/?o=2.1merge-tags/` you stay on the listing page.

This also affects paging querystrings.

**Fix**

Use `request.path` instead of `request.get_full_path()`.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,7 +6,7 @@ Changelog
 
 * Add an admin command to remove orphaned tags
 * Remove support for Python 3.8
-* Replace `request.get_full_path()` with `request.path` for merge form action
+* Fix an issue where the admin merge tag form redirect would fail when querystrings are present inside the URL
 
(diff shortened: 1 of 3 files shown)


rtpg:
Thanks for diagnosing this! Just changed this up a bit to call `reverse` directly instead of manually building up the URL, @garrettc would you be able to sanity check the changes I pushed up?

garrettc:
@rtpg Looks good to me.

rtpg:
alrighty, merging this! Thanks again

PR #947, opened 2026-08-11
Add TaggableManager.remove_by_slug()

### LINKED ISSUE #151: Enhacement: Add remove_by_slug() method to TaggableManager
That way it would also be possible to remove tags by providing their unique slugs. Have in mind that people might want to use URLs like `.../some-object/tags/my-precious-tag/remove/` to remove a tag.

Adding this method seems easy, but I am reluctant to contribute a PR because I am not sure if I can write correct tests for this code addition. If some guidance and supervision is provided I am willing to try.

juneja-varun:
Closes #151.

`remove()` only matches tags by name, so there was no way to remove a tag by slug - useful for endpoints like `.../tags/<slug>/remove/` where the slug is what's actually in the URL, not the display name.

Adds `TaggableManager.remove_by_slug(*slugs)`, mirroring `remove(*tags)` but filtering on `tag__slug` instead of `tag__name`. Refactored `remove()`'s body into a small `_remove_by(**lookup)` helper shared by both methods rather than duplicating the `m2m_changed` signal handling.

Added tests covering basic removal, the no-op case, signals, and that matching is by slug and not name. Docs and changelog updated to match.


juneja-varun:
Just checking in on this one — happy to make any changes if useful, or let me know if you'd rather take a different approach.

